PR #2, opened 2019-01-09
Basic implementations for stream and context for Cubeb

ChunMinChang:
Implement a dummy [*cubeb*][cubeb] stream and a dummy *cubeb* context and add tests for them. Those tests are called via [*cubeb-rs*][cubeb-rs], which is the *Rust* interface for *cubeb* implementations on different platforms.

[cubeb]: https://github.com/kinetiknz/cubeb "kinetiknz/cubeb"
[cubeb-rs]: https://github.com/djg/cubeb-rs "djg/cubeb-rs"

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -67,9 +67,14 @@ impl ContextOps for AudioUnitContext {
         _latency_frame: u32,
         _data_callback: ffi::cubeb_data_callback,
         _state_callback: ffi::cubeb_state_callback,
-        _user_ptr: *mut c_void,
+        user_ptr: *mut c_void,
     ) -> Result<Stream> {
-        Err(Error::not_supported())
+        // The stream must be boxed since capi_stream_destroy releases the
+        // stream by Box::from_raw.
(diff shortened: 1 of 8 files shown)


src/backend/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+        // The context must be boxed since capi_destroy releases the context
+        // by Box::from_raw.
+        let ctx = Box::new(AudioUnitContext {
+            ops: &OPS as *const Ops,
+        });
+        Ok(unsafe { Context::from_ptr(Box::into_raw(ctx) as *mut ffi::cubeb) })
+    }
+
+    fn backend_id(&mut self) -> &'static CStr {
+        unsafe { CStr::from_ptr(b"audiounit\0".as_ptr() as *const c_char) }
```
achronop: We could call it "audiounit-rust" to differentiate from the c backend.

ChunMinChang: Thanks for the catch! I'll upload a commit to address it!

PR #3, opened 2019-07-31
Calculate input_frames of duplex stream after counting missing frames

ChunMinChang:
The frame length to the resampler will be incorrect if the missing
frames are not counted. This solves the crash on [BMO 1570077.](https://bugzilla.mozilla.org/show_bug.cgi?id=1570077)


ChunMinChang:
The changes are reviewed in #4. Carry the approval here and merge the changes to master.

PR #5, opened 2019-07-31
Port of the fix made in the C++ backend to the Rust backend

padenot:


src/backend/mod.rs
```diff
@@ -595,6 +617,17 @@ extern "C" fn audiounit_output_callback(
         slice::from_raw_parts_mut(ptr, len)
     };
 
+    let now = host_time_to_ns(unsafe { mach_absolute_time() });
+    let audio_output_time = host_time_to_ns(unsafe {(*tstamp).mHostTime});
+    let output_latency_ns = (audio_output_time - now) as u64;
```
ChunMinChang: Not sure why using `audio_output_time - now`.

Is `audio_output_time` a future time that starts consuming the output buffer?

src/backend/mod.rs
```diff
@@ -595,6 +617,17 @@ extern "C" fn audiounit_output_callback(
...
+
+    const NS2S: u64 = 1_000_000_000;
+    // The total output latency is the timestamp difference + the stream latency +
+    // the hardware latency.
+    let out_hw_rate = stm.core_stream_data.output_hw_rate as u64;
+    let out_latency: u32 = (output_latency_ns * out_hw_rate / NS2S + stm.current_latency_frames.load(Ordering::SeqCst) as u64) as u32;
+    stm.total_output_latency_frames.store(out_latency, Ordering::SeqCst);
+
```
ChunMinChang, line 633: If the time of function call can be ignored, I prefer merging these code into a function.

ChunMinChang (approved):
Overall looks good. 
Please run `Cargo fmt` to make it looks rustier.

src/backend/mod.rs
padenot: Yes. `mHostTime` is the time at which the first sample of the buffer will be output by the hardware, and it's in the future.

PR #6, opened 2019-08-02
Simplify the APIs to get the device info from hardware

ChunMinChang:
Create several functions to get the device info separately instead of getting all the info in one huge function. As a result, we don't need to call this huge function when we only need one specific device value and it's easier to test the device info query separately.


ChunMinChang:
@kinetiknz: Paul is on PTO, can you review this ?

kinetiknz:
> @kinetiknz: Paul is on PTO, can you review this ?

No problem.  Will take a look ASAP.

kinetiknz:
r+, LGTM!

PR #8, opened 2019-08-16
Simplify the APIs for querying device information

ChunMinChang:
Move most of the APIs that queries the device information to a device-API module so the logic for the main cubeb code (_mod.rs_) will be simpler, without knowing the detail of how the API works. In addition, those APIs are simpified by using other low-level API wrappers.

r? @kinetiknz


kinetiknz:
r+, LGTM!

PR #13, opened 2019-09-13
Change the policy of aggregate usage

ChunMinChang:
In some case, the cubeb stream will use an unexpected mic rather than the one we choose. The STR can be found in [BMO 1563475](https://bugzilla.mozilla.org/show_bug.cgi?id=1563475). That is because we use aggregate device. If the output device is a headset with mic, when we select another mic as the input device, we will create an aggregate device as the input and output devices in the cubeb stream. However, the main mic is likely to be set to the mic within the headset, which is an unexpected choice for users. We should change our policy to use aggregate device when

1. the input device and output device are different
2. the input device is a microphone-only device
3. the output device is a speaker-only device

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -2433,11 +2433,7 @@ impl<'ctx> CoreStreamData<'ctx> {
         self.output_stream_params.rate() > 0
     }
 
-    fn should_use_aggregate_device(
-        &self,
-        input_device: &device_info,
-        output_device: &device_info,
-    ) -> bool {
+    fn should_use_aggregate_device(&self) -> bool {
         // Only using aggregate device when
         // 1. Stream is duplex
         // 2. Input device is valid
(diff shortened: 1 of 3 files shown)


ChunMinChang:
@padenot: I am not sure if we need to use the [ringbuf](https://bugzilla.mozilla.org/show_bug.cgi?id=1563475#c4) like what we did in https://github.com/djg/cubeb-pulse-rs/pull/42. It looks like `input_linear_buffer` in the current implementation. Do I misunderstand something?

padenot:
`input_linear_buffer` works, but uses a lock iirc. `ringbuf` is lock and wait free.

padenot (approved):
Looks good, thanks. We need to test it thoroughly now.

ChunMinChang:
> `input_linear_buffer` works, but uses a lock iirc. `ringbuf` is lock and wait free.

It's actually a wrapper of [`Vec`](https://github.com/ChunMinChang/cubeb-coreaudio-rs/blob/3cf148f341a2dcbcca78bd19dab18a87308f7c70/src/backend/auto_array.rs#L18). But using _ringbuf_ can reduce the maining effort for `input_linear_buffer` I guess.

padenot:
We've discussed on a call, it's better to still allow aggregate device in the common case where input and output is on the same device.

ChunMinChang:
> We've discussed on a call, it's better to still allow aggregate device in the common case where input and output is on the same device.

Does that mean we should keep using aggregate device even if the speaker or microphone is an in-out device? Don't check the microphone is input-only and speaker is output-only?


I think it's worth mentioning there are two cases  that can be called _same device_:

1. The input `AudioObjectID` is __different__ from output `AudioObjectID`. One physical device that has different `AudioObjectID` for input and output, e.g., _AirPods_.
2. The input `AudioObjectID` is __same__ from output `AudioObjectID`, e.g., _plantronics 628_

Given a duplex stream, the original code, without the current change, will only use aggregate device for case 1, not matter input/output device is input/output or in-out.

With the the current change, our code will use aggregate device when all the following conditions are true:
1. the mic device is input-only 
2. the speaker device is output only
3. input `AudioObjectID` is __different__ from output `AudioObjectID`. 

We will keep using aggregate device for case 1. Both the original code and this patch won't support the case 2.

padenot:
It might be better to think in terms of result: for a user, it should "work" in all cases, regardless of what input or output device is selected on what device: input should be from the input device selected, output from the output device selected. What do you think would be the preferred way of doing this?

ChunMinChang:
Hmm.... the perfect way to keep using the aggregate device in common case is to find a way to set the main microphone or main speaker, so the main microphone or main speaker could be the device selected by users. But I don't know how to do that now (I will need to do some research).

If there is no way to do it, we should compare the priority of "always using the device selected by users" and the priority of "keeping the current sound quality" (no drift compensation if we don't use aggregate device). Not sure if it's possible to add telemetry in cubeb. It would be helpful if we can know if how many users that using different devices for input and output, and one of the is in-out.

padenot:
Ok, for now, let's land something where the right devices are always in use, and see if it really causes issues with sound quality.

ChunMinChang:
Ok, I am going to merge this and see if it causes issues. It's easy to revert this anyway.

PR #14, opened 2019-09-13
Improve device switch tests

ChunMinChang:
1. The current device's index of the cached devices will be calculated every time when the DeviceSwitcher.next is called. We should calculate it once the DeviceSwitcher is created and cache it. However, it will also add a limit that devices are not allowed to be changed during the device-changing test since they are cached and assumed to be valid during the whole test.
2. Show the available device list before rotating/switching them is extremely useful for debugging. When the aggregate device is created accidentally in the stream reinitialization when I was working in https://github.com/ChunMinChang/cubeb-coreaudio-rs/pull/13, the `test_switch_device` fails. The reason is that the created aggregate device will be cached as one of the available candidates of the default device. However, it will be destroyed already when it's set to the default device. Showing all the cached devices is useful to know whether we cache some invalid devices that would be assigned to default devices.

### DIFF
--- a/src/backend/tests/api.rs
+++ b/src/backend/tests/api.rs
@@ -31,14 +31,6 @@ fn test_make_sized_audio_channel_layout_with_wrong_size() {
     let _ = make_sized_audio_channel_layout(wrong_size);
 }
 
-// has_input
-// ------------------------------------
-// TODO
-
-// has_output
-// ------------------------------------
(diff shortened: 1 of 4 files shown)


ChunMinChang:
I found I forget to ask review this PR.

padenot (approved):

PR #15, opened 2019-09-30
Fix draining

padenot:
Same fix as https://github.com/kinetiknz/cubeb/pull/546. This only seem to matter in practice when remoting is enabled, because the buffers are not zeroed out.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -756,18 +756,19 @@ extern "C" fn audiounit_output_callback(
         // Post process output samples.
         if stm.draining.load(Ordering::SeqCst) {
             // Clear missing frames (silence)
-            let count_bytes = |frames: usize| -> usize {
+            let frames_to_bytes = |frames: usize| -> usize {
                 let sample_size =
                     cubeb_sample_size(stm.core_stream_data.output_stream_params.format());
-                frames * sample_size / mem::size_of::<u8>()
+                let channel_count = stm.core_stream_data.output_desc.mChannelsPerFrame as usize;
+                frames * sample_size * channel_count
             };
             let out_bytes = unsafe {
                 slice::from_raw_parts_mut(
                     output_buffer as *mut u8,
-                    count_bytes(output_frames as usize),
+                    frames_to_bytes(output_frames as usize),
                 )
             };
-            let start = count_bytes(outframes as usize);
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -759,7 +759,7 @@ extern "C" fn audiounit_output_callback(
             let frames_to_bytes = |frames: usize| -> usize {
                 let sample_size =
                     cubeb_sample_size(stm.core_stream_data.output_stream_params.format());
-                let channel_count = stm.core_stream_data.output_desc.mChannelsPerFrame as usize;
+                let channel_count = stm.core_stream_data.output_stream_params.channels() as usize;
```
ChunMinChang, line 762: Why `output_desc.mChannelsPerFrame` is replaced by `output_stream_params.channels()`?

padenot: For the same reason as https://github.com/kinetiknz/cubeb/commit/fcd9a380e0b0ae58d601a07338ef7f9569b046bd, we're before the mixer here, so we need to take into account the cubeb stream channel count, not the audiounit stream channel count.

ChunMinChang: Oh, I think I got it. Just double check:
1. if `output_stream_params.channels()` is __same__ from the `output_desc.mChannelsPerFrame`
    - mixer is `None`
    - `output_buffer` here is `buffers[0].mData` from callback
2. if `output_stream_params.channels()` is __different__ from the `output_desc.mChannelsPerFrame`
    - mixer is `Some`thing whose internal `channel` value is `output_stream_params.channels()`
    - `output_buffer` here is a pointer pointing to the internal buffer of the mixer
    - the internal buffer size is based on its internal `channel`, which is `output_stream_params.channels()`

For the case 1, the logic is straightforward. For the case 2, when `output_desc.mChannelsPerFrame` > `output_stream_params.channels()`, using `output_desc.mChannelsPerFrame` will touch the invalid address that is out of the bound of mixer's internal buffer.

This makes me think we should expose the `channel` value from mixer, or create a `zero_buffer` in the mixer.

ChunMinChang (approved):
The mystery is solved! \o/

src/backend/mod.rs
padenot: Yes, exactly.

PR #20, opened 2019-10-18
Make utf8_from_cfstringref work with empty CFStringRef

ChunMinChang:
Passing a `CFStringRef `whose internal string is `""` (empty) to
`StringRef::utf8_from_cfstringref` will hit an assertion that asserts the passed
`CFStringRef` cannot be empty. Instead of using the assertion, an empty `u8`
`Vec` should be returned directly when the passed `CFStringRef` is empty,
which will yield an empty `String` or `CString` in `StringRef::to_string` or `StringRef::to_cstring`.

### DIFF
--- a/coreaudio-sys-utils/src/string.rs
+++ b/coreaudio-sys-utils/src/string.rs
@@ -81,7 +81,9 @@ fn utf8_from_cfstringref(string_ref: CFStringRef) -> Vec<u8> {
     assert!(!string_ref.is_null());
 
     let length: CFIndex = unsafe { CFStringGetLength(string_ref) };
-    assert!(length > 0);
+    if length == 0 {
+        return Vec::new();
+    }
 
     // Get the buffer size of the string.
     let range: CFRange = CFRange {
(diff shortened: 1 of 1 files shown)


padenot (approved):
Do you think this could be the cause for https://bugzilla.mozilla.org/show_bug.cgi?id=1589931 ?

ChunMinChang:
> Do you think this could be the cause for https://bugzilla.mozilla.org/show_bug.cgi?id=1589931 ?

hmm...I think it's not. `current_device` doesn't use `CFStringRef`.

https://github.com/ChunMinChang/cubeb-coreaudio-rs/blob/a39bf5f71fff3a46d7ee4e2cdbb5770e49d2d40c/src/backend/mod.rs#L3461-L3470

https://github.com/ChunMinChang/cubeb-coreaudio-rs/blob/a39bf5f71fff3a46d7ee4e2cdbb5770e49d2d40c/src/backend/mod.rs#L1461-L1468

The `device.output_name` is `source.into_raw()`, where `source` is a `CString`, if there is a valid output device. Otherwise `device.output_name` is `nullptr` by default. IIRC, passing a `nullptr` to `strcmp` causes undefined behavior. Maybe we can test this by running `sudo kextunload ...` to drop the output devices.

ChunMinChang:
> Do you think this could be the cause for https://bugzilla.mozilla.org/show_bug.cgi?id=1589931 ?

We can fix this in another pull request. I am going to merge this one.

PR #22, opened 2019-10-24
Update dependencies

ChunMinChang:
- Remove unused _core-foundation-sys_ dependency. The _core-foundation-sys_ is imported for `CFStringRef` previously, which is wrapped into a `StringRef` in _coreaudio-sys-utils_.
- Bump _coreaudio-sys_ to 0.2.3 that fixes the cross-compiling issue.

### DIFF
--- a/coreaudio-sys-utils/Cargo.toml
+++ b/coreaudio-sys-utils/Cargo.toml
@@ -10,4 +10,4 @@ core-foundation-sys = { version = "0.6" }
 [dependencies.coreaudio-sys]
 default-features = false
 features = ["audio_unit", "core_audio"]
-version = "0.2.3"
+version = "0.2"
--- a/coreaudio-sys-utils/Cargo.toml
+++ b/coreaudio-sys-utils/Cargo.toml
@@ -10,4 +10,4 @@ core-foundation-sys = { version = "0.6" }
 [dependencies.coreaudio-sys]
 default-features = false
 features = ["audio_unit", "core_audio"]
-version = "0.2"
+version = "0.2.3"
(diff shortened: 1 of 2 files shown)


coreaudio-sys-utils/Cargo.toml
```diff
@@ -10,4 +10,4 @@ core-foundation-sys = { version = "0.6" }
 [dependencies.coreaudio-sys]
 default-features = false
 features = ["audio_unit", "core_audio"]
-version = "0.2"
+version = "0.2.3"
```
glandium: This is not necessary. You'll get the newer 0.2 version with cargo update -p coreaudio-sys if you remove the patch.crates-io.coreaudio-sys section in Cargo.toml in mozilla-central. (you need to run mach vendor rust after changing Cargo.toml, though)

ChunMinChang: If someone already has a 0.2.2 version, will it be rebuilt?

glandium: Outside of mozilla-central, not until they'd run cargo update.

glandium: That said, they wouldn't get the new version of cubeb-coreaudio-rs without running cargo update either.

glandium: Except if they are building cubeb-coreaudio-rs directly, rather than depend on it.

ChunMinChang: I don't think it's used outside of Mozilla so it can be removed.

ChunMinChang: Thanks for the review!

padenot (approved):

PR #23, opened 2019-11-04
Implement a new any-to-any mixer

ChunMinChang:
Create an oxidized cubeb-mixer, which is adapted from [cubeb_mixer.cpp][0], as the new mixer module. The new mixer is able to mix from any input channels to any output channels, without predefined channel order.

The mixer would mix audio input data to audio output data based on the given input and output audio _channel layout_. There are at least 2 different concepts when refering to audio _channel layout_.
1. _channel count_: number of channels
2. _channel order_: the order of the audio channel

In the past, we have the following issues when mixing the audio
1. The _channel count_ and the _channel order_ of a _channel layout_ are predefined. In our current implementation, for a _stereo_ layout, its _channel count_ must be _2_ and its _channel order_ must be _[front-left, front-right]_. However, in macOS, users are allowed to configure an _N_-channel device to any _1-to-N_ channel device. For example, a _6_-channel output device can be configured to a _stereo_ device that only uses _2_ channels and the left _4_ channels should be _ignored/unused_. That is, a _stereo channel layout_ whose _channel count_ could be _6_ and its _channel order_ could be _[unused, front-right, unused, unused, front-left, unused]_.
2. If the _channel count_ of a _channel layout_ is different from the predefined one, the _channel layout_ would be marked as an invalid layout and the audio data would be [passed through from input buffer to output buffer][1], without redirecting the audio data to its mapping channel. For example, the data of indices _0_ and _1_ in the audio input buffer whose layout is _stereo: [front-left, front-right]_ should be copied as the data of indices _3_ and _1_  in the audio output buffer whose layout _stereo: [unused, front-right, unused, front-left]_.
3. If the _channel count_ of a _channel layout_ is the same as the predefined one, the current mixer would ignore the user-defined _channel order_ and mix the audio data by the predefined channel order. For example, a _stereo_ output with 2 channels would always be treated as its channel order is _[front-left, front-right]_. However, users are allowed to set _stereo_ output to _[front-right, front-left]_ in macOS.

1. We would ignore the user-defined channel layout and force the output device to apply our pre-defined channel layout, which is a standard _SMPTE_ layout
    https://github.com/ChunMinChang/cubeb-coreaudio-rs/blob/cd954ca94baafc1bd0c4c661f07b816e9263ed1b/src/backend/mod.rs#L2722-L2731

The new mixer is created to fix the above 1-3 problems. The core idea is to think the mixing as matrix multiplication. The input and output audio data, _in_audio_, _out_audio_, are _Mx1_, _Nx1_ matrix and the mixing coefficients are a _NxM_ matrix _m_ such that _out_audio = m x in_audio_

```
data in channel #1 ▸ │ Silence    │   │ 0, 0, 0, 0 │   │ FrontRight   │ ◂ data in channel #1
data in channel #2 ▸ │ FrontRight │ = │ 1, C, 0, L │ x │ FrontCenter  │ ◂ data in channel #2
data in channel #3 ▸ │ FrontLeft  │   │ 0, C, 1, L │   │ FrontLeft    │ ◂ data in channel #3
                            ▴                 ▴        │ LowFrequency │ ◂ data in channel #4
                            ┊                 ┊                ▴
                            ┊                 ┊                ┊
                        out_audio      mixing matrix m       in_audio

where C, L are coefficients for mixing FrontCenter and LowFrequency data 
to FrontLeft or FrontRight channels. The Silence channel is a unused channel in the output device, so its channel data would be zero.
```

The mixing coefficient matrix _m_ would be generated dynamically, based on given _in_audio_ and _out_audio_. As a result, the restriction of the channel order would be loosen and audio is able to be mixed from any channel layout to any channel layout, without any pre-defined channel order and limitation of channels' number.

[0]: https://github.com/kinetiknz/cubeb/blob/81432070256d4e536cfcac14fe2b90f13866f369/src/cubeb_mixer.cpp
[1]: https://github.com/kinetiknz/cubeb/blob/81432070256d4e536cfcac14fe2b90f13866f369/src/cubeb_mixer.cpp#L570-L584


ChunMinChang:
I would probably make this new mixer as an independent crate so we could reuse it in Windows. But I'll do in another pull request.

mixer/src/coefficient.rs
```diff
@@ -0,0 +1,689 @@
...
+    // │ FrontCenter  │ = M x │ FrontCenter  │ ◂ data in front-center channel
+    // │ ...........  │       │ ...........  │ ◂ ...
+    // │ Silence      │       │ Silence      │ ◂ data in silence channel
+    //        ▴                       ▴
+    //    out_audio               in_audio
+    //
+    // That is,
+    // 1. If the input-channel is slience, it won't be mixed into any channel.
+    // 2. If the output-channel is slience, its output-channel data will be zero (slience).
+    // 3. If input-channel j is different from output-channel i, audio data in input channel j
```
padenot, line 358: `s/slience/silence/` a few times.

mixer/src/coefficient.rs
```diff
@@ -0,0 +1,689 @@
...
+
+impl MixingCoefficient for i16 {
+    type Coef = i32;
+
+    fn max_coefficients_sum() -> f64 {
+        1.0
+    }
+
+    fn coefficient_from_f64(value: f64) -> Self::Coef {
+        // 32767 = 0x8000 = 2^15
```
padenot: Write that then, no need to comment and hard code the real value. `(2<<15)` is meaningful enough.

src/backend/mixer.rs
```diff
@@ -1,53 +1,214 @@
...
+        // Bose QC35, mark 1 and 2), expose a single channel mapped to the right for
+        // some reason.
+        // TODO: Only apply this setting when device is Bose QC35 (by device_property.rs).
+        if out_channel_count == 1 {
+            output_channels = vec![mixer::Channel::FrontCenter];
+        } else if out_channel_count == 2 {
+            output_channels = vec![mixer::Channel::FrontLeft, mixer::Channel::FrontRight];
+        }
+
+        let all_slience = vec![mixer::Channel::Silence; out_channel_count];
```
padenot: more typos

mixer/src/coefficient.rs
```diff
@@ -0,0 +1,689 @@
...
+    pub fn create(input_channels: Vec<Channel>, output_channels: Vec<Channel>) -> Self {
+        let input_layout = ChannelLayout::new(input_channels).expect("Invalid input layout");
+        let output_layout = ChannelLayout::new(output_channels).expect("Invalid output layout");
+
+        let mixing_matrix =
+            Self::build_mixing_matrix(input_layout.channel_map, output_layout.channel_map)
+                .unwrap_or_else(|_| Self::get_basic_matrix());
+
+        // TODO: Create a NxM mixing matrix directly rather than creating a CHANNELSxCHANNELS
+        // matrix first than picking the used coefficients.
```
padenot: Why ?

mixer/src/coefficient.rs
```diff
@@ -0,0 +1,689 @@
...
+        let coefficient_matrix = Self::pick_coefficients(
+            &input_layout.channels,
+            &output_layout.channels,
+            &mixing_matrix,
+        );
+
+        // TODO: No need to normalize the coefficient_matrix when it's from get_basic_matrix.
+        // Since no duplicate channels in input, in each row, only one column is 1 and all other
+        // columns are all 0.
```
padenot: Why add another path here, the normal path works? It's doing a little bit more works, but we're only building the matrix once, limiting code complexity is more important.

mixer/src/coefficient.rs
```diff
@@ -0,0 +1,689 @@
...
+            }
+        }
+        coefficients
+    }
+}
+
+pub trait MixingCoefficient {
+    type Coef;
+
+    // TODO: These should be private.
```
padenot, line 417: Do it?

padenot (commented):
This should generally works. I'm interested in two things:
- Do the inner loop optimizes well? There are function calls, etc. but the optimizer should see through that. I'm hoping it can succeed in removing the bound checks, but I'm not sure since this is using array indexing.
- What do you think is the common case? To me, a matrixing multichannel mixer is more of an application concern and not really an audio IO concern. You take the file or the application, and its output channel count and mapping, then you have a look at what the audio system can do. Then if there are mismatch, we can rematrix.

For example, a 5.1 file to a 5.1 system should be pass-through, using the OS facility to do that, but a 5.1 to a 2.0 system should be rematrixed. A 2.0 file on a 5.1 system should be trying to output stereo via the OS, and not rematrix.

mixer/src/coefficient.rs
ChunMinChang: Picking-coefficients, `Self::pick_coefficients`, is a process of copying the coefficients from the temporary CHANNELSxCHANNELS mixing matrix to an un-normalized NxM mixing matrix, where _M, N_ are input channel count and output channel count.

Instead of creating a temporary mixing matrix then copying the coefficients, we could build the un-normalized NxM mixing matrix directly.

Assume we have a 1-channel _mono_ audio file on input side and a 4-channel speaker with stereo layout on the output side. The input channels are _[front-center]_ and the output channels are _[unused, front-right, unused, front-left]_.

In the current implementation, we would create a basic _19 x 19_ mixing matrix by `get_basic_matrix()` first, and then set the coefficients in `matrix[FRONT_LEFT][FRONT_CENTER]` and `matrix[FRONT_RIGHT][FRONT_CENTER]`:

```
matrix[FRONT_LEFT][FRONT_CENTER] += coefficient;   // matrix[0][2] = coefficient
matrix[FRONT_RIGHT][FRONT_CENTER] += coefficient;  // matrix[1][2] = coefficient
```
where the `FRONT_LEFT`, `FRONT_RIGHT`, `FRONT_CENTER` are the indices _0, 1, 2_ in the _19 x 19_ mixing matrix, which maps to predefined value in `enum Channel`. 

Next, we would create _4 x 1_ matrix _M_

```
| 0           | // unused     , M[0][0] from matrix[SILENCE][FRONT_CENTER]
| coefficient | // front-right, M[1][0] from matrix[FRONT_RIGHT][FRONT_CENTER]
| 0           | // unused     , M[2][0] from matrix[SILENCE][FRONT_CENTER]
| coefficient | // front-left , M[3][0] from matrix[FRONT_LEFT][FRONT_CENTER]
```
where `M[1][0]` is copied from `matrix[FRONT_RIGHT][FRONT_CENTER] = matrix[1][2]`, 
`M[3][0]` is copied from `matrix[FRONT_LEFT][FRONT_CENTER] = matrix[0][2]` and all else are from `matrix[SILENCE][FRONT_CENTER] = matrix[18][2]`.

Actually, the above process could be simplified if `FRONT_LEFT`, `FRONT_RIGHT`, `FRONT_CENTER` can be calculated dynamically to indicate the indices of _front-left_, _front-right_ channel in the output buffer and the index of _front-center_ channel in the input buffer. That is, if `FRONT_LEFT`, `FRONT_RIGHT`, `FRONT_CENTER` can be defined to `3`, `1` and `0` then we can create a _4 x 1_ `matrix` whose initial values are all `0` and do
```
matrix[FRONT_LEFT][FRONT_CENTER] += coefficient;   // matrix[3][0] = coefficient
matrix[FRONT_RIGHT][FRONT_CENTER] += coefficient;  // matrix[1][0] = coefficient
```
when setting the coefficients for mixing input's front-center to output's front-stereo.

mixer/src/coefficient.rs
ChunMinChang: Yeah, the normal path works and the the preprocessing only works once so it should be fine. I'll remove the comments.

mixer/src/coefficient.rs
ChunMinChang: I don't have a good way to deal with it actually. If I put

```rust
trait X {
    fn max_coefficients_sum() -> f64; // Used for normalizing.
    fn coefficient_from_f64(value: f64) -> Self::Coef;
    fn would_overflow_from_coefficient_value(coefficient: &[Vec<f64>]) -> Option<bool>;
}
```
In another trait `X`, then I would need to expose `X` (`X` becomes `pub`)

```rust
pub struct Coefficient<T>
where
    T: MixingCoefficient + X, // X is exposed
    T::Coef: Copy, 
```

The other issue is `Self::Coef`. The output of `coefficient_from_f64` should be `MixingCoefficient::Coef` but I didn't find the way to set it properly...

ChunMinChang:
> This should generally works. I'm interested in two things:
> 
>     * Do the inner loop optimizes well? There are function calls, etc. but the optimizer should see through that. I'm hoping it can succeed in removing the bound checks, but I'm not sure since this is using array indexing.

I dumped the MIR by `cargo rustc --lib -- --emit mir` but I barely recognize it. I'll use some benchmark crate like _criterion_ to does some tests. We used to have the special handling for some (common) channel count: https://github.com/kinetiknz/cubeb/blob/81432070256d4e536cfcac14fe2b90f13866f369/src/cubeb_mixer.cpp#L475. I'll try to use special handling to see if the special handling really matters, by the third-party benchmark crate.


>     * What do you think is the common case? To me, a matrixing multichannel mixer is more of an application concern and not really an audio IO concern. You take the file or the application, and its output channel count and mapping, then you have a look at what the audio system can do. Then if there are mismatch, we can rematrix.
> 
> 
> For example, a 5.1 file to a 5.1 system should be pass-through, using the OS facility to do that, but a 5.1 to a 2.0 system should be rematrixed. A 2.0 file on a 5.1 system should be trying to output stereo via the OS, and not rematrix.

I guess the most common case is both input and output are stereos (but I never saw the telemetry). In that case, we won't create the mixer. We will only create a mixer when the input channel layout is different from the output channel layout. Is this the problem you worry about?
https://github.com/ChunMinChang/cubeb-coreaudio-rs/blob/0920240e4166d2b562840c8062e149d63f7c3a02/src/backend/mod.rs#L2733-L2754

mixer/src/coefficient.rs
padenot: Yes, but this is only once during initialization, right? Having simple and understandable code is more important isn't it?

mixer/src/coefficient.rs
padenot: Don't worry too much about it.

padenot:
> We will only create a mixer when the input channel layout is different from the output channel layout. Is this the problem you worry about?

We should try to not create the mixer if the input is stereo and the output is more than stereo, and instead try to open a stereo stream on AudioUnit, that will decide on the mappping. This is to solve bugs like bug 1468381.

ChunMinChang:
> I'll use some benchmark crate like _criterion_ to does some tests. We used to have the special handling for some (common) channel count: https://github.com/kinetiknz/cubeb/blob/81432070256d4e536cfcac14fe2b90f13866f369/src/cubeb_mixer.cpp#L475. I'll try to use special handling to see if the special handling really matters, by the third-party benchmark crate.

I've introduced the _criterion_ crate for the benchmark. Now, the benchmark test could be run by `cargo bench` (There would be a plot report at _target/criterion/report/index.html_ if _gnuplot_ is installed).

The code with special handling was compared with the original implementation, in 2 scenarios with `f32`, `i16` types:
- downmix from 5.1 to stereo
- upmix from mono to stereo

In short, with the special handling (code below), the `mix` would be slightly faster. When downmixing the audio 5.1 to the stereo with `512` _frames_, the special handling would be faster about _1 us_.

But I am not sure if it's worth adding complexity to gain this improvement. What do you think?

## Benchmark result

### Without special handling
---------
downmix_f32             time:   [19.276 us 19.345 us 19.419 us]                         
                        change: [+0.0700% +0.8003% +1.6345%] (p = 0.04 < 0.05)
                        Change within noise threshold.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low mild
  4 (4.00%) high mild
  4 (4.00%) high severe

downmix_i16             time:   [19.263 us 19.507 us 19.780 us]                         
                        change: [-1.8844% -0.2238% +1.5090%] (p = 0.79 > 0.05)
                        No change in performance detected.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe

upmix_f32               time:   [8.2549 us 8.2852 us 8.3193 us]                       
                        change: [-1.8970% -1.0369% -0.1848%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

upmix_i16               time:   [9.2803 us 9.3130 us 9.3497 us]                       
                        change: [-0.1100% +0.8325% +1.6188%] (p = 0.06 > 0.05)
                        No change in performance detected.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe

### With special handling
-------------
downmix_f32             time:   [18.069 us 18.184 us 18.301 us]                         
                        change: [-0.2277% +0.4292% +1.0855%] (p = 0.21 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

downmix_i16             time:   [18.166 us 18.291 us 18.421 us]                         
                        change: [-1.2536% -0.3453% +0.5522%] (p = 0.45 > 0.05)
                        No change in performance detected.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe

upmix_f32               time:   [8.1000 us 8.1291 us 8.1582 us]                       
                        change: [+0.2135% +0.9894% +1.7244%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) low mild
  1 (1.00%) high mild
  2 (2.00%) high severe

upmix_i16               time:   [9.0213 us 9.0667 us 9.1131 us]                       
                        change: [-0.9681% -0.3106% +0.3800%] (p = 0.36 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

## Code with special handling for common input channel count
```rust
#[derive(Debug)]
pub struct Mixer<T>
where
    T: Copy + Debug + MixingCoefficient,
    T::Coef: AddAssign
        + Copy
        + Debug
        + Default
        + Mul<T::Coef, Output = T::Coef>
        + Add<T::Coef, Output = T::Coef>,
{
    coefficient: Coefficient<T>,
    coef_indices: Vec<Vec<usize>>,
}

impl<T> Mixer<T>
    ....

    pub fn new(input_channels: Vec<Channel>, output_channels: Vec<Channel>) -> Self {
        let coefficient = Coefficient::create(input_channels, output_channels);
        let mut coef_indices = Vec::new();
        for (i, _) in coefficient.output_channels().iter().enumerate() {
            let mut indices = Vec::new();
            for (j, _) in coefficient.input_channels().iter().enumerate() {
                if coefficient.get(j, i) != T::Coef::default() {
                    indices.push(j);
                }
            }
            coef_indices.push(indices);
        }
        Self {
            coefficient,
            coef_indices,
        }
    }

    pub fn mix(&self, input_buffer: &[T], output_buffer: &mut [T]) {
        assert_eq!(
            input_buffer.len(),
            self.input_channels().len(),
            "input slice must have the same size as the input channel's one."
        );
        assert_eq!(
            output_buffer.len(),
            self.output_channels().len(),
            "output slice must have the same size as the output channel's one."
        );
        for (i, output) in output_buffer.iter_mut().enumerate() {
            // T must implement Default that returns a zero value from default().
            let mut value = T::Coef::default(); // Create a zero value.

            match self.coef_indices[i].len() {
                0 => {
                    // do nothing
                }
                1 => {
                    let idx = self.coef_indices[i][0];
                    value =
                        T::to_coefficient_value(input_buffer[idx]) * self.coefficient.get(idx, i);
                }
                2 => {
                    let idx1 = self.coef_indices[i][0];
                    let idx2 = self.coef_indices[i][1];
                    value = T::to_coefficient_value(input_buffer[idx1])
                        * self.coefficient.get(idx1, i)
                        + T::to_coefficient_value(input_buffer[idx2])
                            * self.coefficient.get(idx2, i);
                }
                n => {
                    for k in 0..n {
                        let idx = self.coef_indices[i][k];
                        value += self.coefficient.get(idx, i)
                            * T::to_coefficient_value(input_buffer[idx]);
                    }
                }
            }
            *output = T::from_coefficient_value(
                value,
                self.coefficient.would_overflow_from_coefficient_value(),
            );
        }
    }

    ....
```

mixer/src/coefficient.rs
ChunMinChang: I thought it would be easier to read the code if I do that... Yes, since the mixing matrix would be done only once during initialization so it wouldn't bring much performance benefit if I implemented the idea above. The choice matters here is for readability.

I'll move those todos in a note document instead.

ChunMinChang:
> This is to solve bugs like bug 1468381.

Applying the current patch can solve Bug 1468381. This patch uses the user-defined layout and the channel data will be redirected to the user-defined channel.

> We should try to not create the mixer if the input is stereo and the output is more than stereo, and instead try to open a stereo stream on AudioUnit, that will decide on the mappping.

I've tried this but it doesn't work in general. Specifically, it only works in one specifical situation:

If the device layout of the input audio file is *A_in: [in_1, in_2, ... , in_k]*, then the mixer can only be skipped when **all** the following conditions are true:
1. the output device layout *A_out* is *[in_1, in_2, ... , in_k, out_l, out_m, ... ]* =  *[A_in | others... ]*. That is, *A_out* is *A_in* with padding channels in the tail.
2. The output buffer data provided from the platform are initialized with `0` (silence).

I will explain this in detail below.

My test device is _StarTech.com 7.1 USB Sound Card_, which has _8_ output channels. The device layout can be set to _stereo_ or _quadraphonic_ or _5.1_ .... etc via _Audio MIDI Setup_. However, no matter what layout the device is set by the user or via the API, the channel count of the device always returns `8` and the output buffer always has an 8-channel size.

For example, if the current (user-defined) device layout is _[back-left, back-right, front-center, low-frequency, front-left, front-right, unused, unused]_ (audio 5.1 in a 8-channel device), device layout can still be set to _[front-left, front-right]_ (standard 2-channel stereo) without any error return. But if I queried the current device layout again, it still returns _[back-left, back-right, front-center, low-frequency, front-left, front-right, unused, unused]_, not _[front-left, front-right]_.
And the size of the output audio buffer is also for _8_-channel, not for _2_-channel.

The size of the output buffer always matches the max channel count of the output device. When we try playing (2-channel) stereo audio, _[front-left, front-right]_, with this 8-channel output device, we need to know its device layout first. If the device layout is stereo that only uses the _3_ and _4_ channels: _[unused, unused, front-left, front-right, unused, unused, unused, unused]_, then we need to map the _front-left_, _front-right_ from _1_ and _2_ channel on input side to the _3_ and _4_ channel on the output side by our own. If the device layout is stereo that only uses the _1_ and _2_ channels: _[front-left, front-right, unused, unused, unused, unused, unused, unused]_, then we probably could skip using the mixer. The output callback that works for _[front-left, front-right]_ layout assuming the buffer is for 2-channel would luckily work when the layout is actually _[front-left, front-right, unused, unused, unused, unused, unused, unused]_, since the channel data would still be fed into the first two-channel. Those unused-channel data in the output buffer wouldn't be touched in this case. If their initial value is `0`, then they are already silenced. It's ok to leave them as they are in this case.

In general case, since the output buffer always matches the max channel count of the output device. We need to get the current (user-defined) layout first and then map the channel data by ourselves.

padenot:
Thanks for the very detailed explanation!

src/backend/mixer.rs
```diff
@@ -1,53 +1,214 @@
...
+        // Check output buffer size.
+        let size_needed = frames * self.output_channels().len() * self.sample_size();
+        assert!(output_buffer_size >= size_needed);
+
+        match self {
+            MixerType::IntegerMixer(m) => {
+                let in_buf_ptr = input_buffer_ptr as *const i16;
+                let out_buf_ptr = output_buffer_ptr as *mut i16;
+                let input_buffer = unsafe {
+                    slice::from_raw_parts(in_buf_ptr, frames * self.input_channels().len())
```
padenot, line 124: In general I'd be happier if we'd have a mixer that is 100% safe rust, and have a little layer on top for backends that pass in raw ptrs and validates the buffer lengths, etc. We can do this in a followup.

padenot (approved):

src/backend/mixer.rs
ChunMinChang: Not sure if I catch this. Let me try: Do you mean it's better to split the `mix` into two parts?
1. Add a `mix` function like `fn mix(..., input_buffer: &[...], output_buffer: &[...], frames: usize)`
2. Add a converter function to convert `buffer_ptr: *mut T` and `buffer_len: usize` into a `&[T]` or `Vec<T>` and feed it into the above `mix`?

I could open an followup for this if the above guess is what you mean.

PR #24, opened 2019-11-12
Add manual tests

ChunMinChang:
There are some crash reports in Bugzilla that relate to stream reinitialization. There are no tests we could test that manually. This change adds some manual tests that loop in the input or output or in-out streams. While the test runs, users could manually unplug or change the current default devices to check if stream reinitialization works. Some cleanup are contained in this change as well.

### DIFF
--- a/coreaudio-sys-utils/src/dispatch.rs
+++ b/coreaudio-sys-utils/src/dispatch.rs
@@ -54,13 +54,13 @@ where
         F: FnOnce(),
     {
         // Retake the leaked closure.
-        let closure: Box<F> = unsafe { Box::from_raw(unboxed_closure as *mut F) };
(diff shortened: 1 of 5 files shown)


coreaudio-sys-utils/src/dispatch.rs
```diff
@@ -54,13 +54,13 @@ where
         F: FnOnce(),
     {
         // Retake the leaked closure.
-        let closure: Box<F> = unsafe { Box::from_raw(unboxed_closure as *mut F) };
+        let closure = unsafe { Box::from_raw(unboxed_closure as *mut F) };
         // Execute the closure.
         (*closure)();
         // closure is released after finishiing this function call.
```
padenot: `s/ii/i/`

padenot (approved):

PR #26, opened 2019-11-26
Avoid various racing issues 

ChunMinChang:
This change solves the racing issues mentioned in [BMO 1598413][b1598413]. Racing can happen when: 
1. stream start and stream destroy runs in parallel
    - The stream data callback may still run when the stream is being destroyed
2. stream start and stream reinit runs in parallel
    - The stream data callback may still run when the stream is being reinitialized
3. stream stop and stream destroy runs in parallel
    - The stream data callback is canceled twice but it's probably fine
4. stream stop and stream reinit runs in parallel
    - The stream-stop may fail silently since reinit may restart stream after stream-stop is called
5. stream destroy and stream reinit runs in parallel
    - The stream-stop may fail silently since reinit may restart stream after stream-stop is called

To avoid the above problems, the stream-start and stream-start should be run in the same task queue where the stream-destroy and stream-reinit tasks are so all these operations are on the same task queue. As a result, they will be executed one by one sequentially, without worrying about racing.

[b1598413]: https://bugzilla.mozilla.org/show_bug.cgi?id=1598413#c1

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -3386,28 +3386,29 @@ impl<'ctx> AudioUnitStream<'ctx> {
         *self.destroy_pending.get_mut() = true;
 
         let queue = self.context.serial_queue;
-        let mutexed_stm = Arc::new(Mutex::new(self));
-        let also_mutexed_stm = Arc::clone(&mutexed_stm);
+
+        let stream_mutex = Arc::new(Mutex::new(self));
+        let stream_clone = Arc::clone(&stream_mutex);
         // Execute close in serial queue to avoid collision
         // with reinit when un/plug devices
         sync_dispatch(queue, move || {
-            let mut stm_guard = also_mutexed_stm.lock().unwrap();
+            let mut stream = stream_clone.lock().unwrap();
 
             // Call stop_audiounits to avoid potential data race. If there is a running data callback,
             // which locks a mutex inside CoreAudio framework, then this call will block the current
             // thread until the callback is finished since this call asks to lock a mutex inside
             // CoreAudio framework that is used by the data callback.
-            if !stm_guard.shutdown.load(Ordering::SeqCst) {
-                stm_guard.core_stream_data.stop_audiounits();
-                *stm_guard.shutdown.get_mut() = true;
+            if !stream.shutdown.load(Ordering::SeqCst) {
+                stream.core_stream_data.stop_audiounits();
+                *stream.shutdown.get_mut() = true;
             }
 
-            stm_guard.destroy_internal();
+            stream.destroy_internal();
         });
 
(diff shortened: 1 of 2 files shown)


ChunMinChang:
try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=f216376090b35e8c526a7a20b3eb2ecfec8b91cd

src/backend/mod.rs
```diff
@@ -3379,30 +3383,32 @@ impl<'ctx> AudioUnitStream<'ctx> {
...
-        }
-
         *self.destroy_pending.get_mut() = true;
 
         let queue = self.context.serial_queue;
-        let mutexed_stm = Arc::new(Mutex::new(self));
-        let also_mutexed_stm = Arc::clone(&mutexed_stm);
+
+        let stream_mutex = Arc::new(Mutex::new(self));
+        let stream_clone = Arc::clone(&stream_mutex);
```
padenot: This is not strictly needed in theory because it's not concurrent, because it's a sync dispatch, right?

And we can't linearize using a lock because CoreAudio itself is using locks and that would cause deadlocks?

padenot (commented):
I need an answer for the question below, but in general it looks like it would fix the bug.

src/backend/mod.rs
ChunMinChang: Yes, the mutex is not really needed since it's sync. 

I've tried using `core_stream_data: Mutex<CoreStreamData<'ctx>>` but I feel it is prone to cause deadlock. For example, the deadlock can be formed when the following two threads run in parallel:
- callback thread holding the mutex inside AudioUnit, and then call `core_stream_data.lock()`
- reinit thread enters a critical section by calling `core_stream_data.lock()` and then try to stop AudioUnit, which tries to lock the mutex inside AudioUnit that is held by callback thread.

I've also tried to add a new member, `lock: Mutex<()>`, in `AudioUnitStream` and use it in the part that should be a critical section, but I hit a twice borrowing issue. 

It probably has a workaround to avoid that borrowing issue, but I think the code would be more clear/simpler if I just dispatch start/stop stuff in the same task queue that `reinit` and `destroy` use, so they would be mutex-free.

padenot: Yeah. But I think what you did works? If you can remove it, do it, otherwise, land it like that. Maybe with a comment explaining what we just said?

padenot (approved):

ChunMinChang:
try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=cc708ccadeac243fca0c1590bd15d8669fdedf6b

src/backend/mod.rs
ChunMinChang: I've updated the code that won't use mutex before `sync_dispatch`.

Issue #31, opened 2019-12-14
intermittent crash when running cargo test


ChunMinChang:
The `cargo test` crashes randomly.

Here is a log for the crash when running `rust-lldb target/debug/deps/cubeb_coreaudio-*`

```
Process 2001 stopped
* thread #24, name = 'backend::tests::api::test_set_channel_layout_input', stop reason = EXC_BAD_ACCESS (code=1, address=0xa)
    frame #0: 0x00007fff3c23e198 CoreFoundation`CFRelease + 11
CoreFoundation`CFRelease:
->  0x7fff3c23e198 <+11>: movq   0x8(%rdi), %rcx
    0x7fff3c23e19c <+15>: movq   (%rdi), %rax
    0x7fff3c23e19f <+18>: testq  %rax, %rax
    0x7fff3c23e1a2 <+21>: je     0x7fff3c23e1cb            ; <+62>
Target 0: (cubeb_coreaudio-fb8fa5520de9e09f) stopped.
(lldb) bt
error: need to add support for DW_TAG_base_type '()' encoded with DW_ATE = 0x7, bit_size = 0
* thread #24, name = 'backend::tests::api::test_set_channel_layout_input', stop reason = EXC_BAD_ACCESS (code=1, address=0xa)
  * frame #0: 0x00007fff3c23e198 CoreFoundation`CFRelease + 11
    frame #1: 0x0000000105bc2134 CoreAudio`___lldb_unnamed_symbol120$$CoreAudio + 34
    frame #2: 0x0000000105c5c300 CoreAudio`___lldb_unnamed_symbol2202$$CoreAudio + 14
    frame #3: 0x0000000105c5cbdf CoreAudio`___lldb_unnamed_symbol2210$$CoreAudio + 37
    frame #4: 0x0000000105be3115 CoreAudio`___lldb_unnamed_symbol672$$CoreAudio + 165
    frame #5: 0x0000000105cbdd2c CoreAudio`___lldb_unnamed_symbol3288$$CoreAudio + 30
    frame #6: 0x00007fff50950653 AudioToolboxCore`AudioComponentInstanceDispose + 58
[...]

PR #32, opened 2019-12-18
Remove the test_set_channel_layout_input

ChunMinChang:
Calling audiounit_get_current_channel_layout by _input-only_
(_microphone-only_) `AudioUnit` is prone to cause crash intermittently.

The `test_set_channel_layout_input` is likely to cause intermittent crash
since it calls `audiounit_get_current_channel_layout` by an _input-only_
`AudioUnit`. Removing this test can avoid the random crashes when running
`cargo test` (#31).

The cubeb implementation has no need to get the channel-layout on input
side, so the `test_set_channel_layout_input` is unnecessary anyway.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -1001,6 +1001,8 @@ fn audiounit_get_preferred_channel_layout(output_unit: AudioUnit) -> ChannelLayo
     audiounit_convert_channel_layout(layout.as_ref())
 }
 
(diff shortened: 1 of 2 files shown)


padenot (approved):
Ok to land this now, but please try to understand the crash.

ChunMinChang:
> Ok to land this now, but please try to understand the crash.

Sure, I'll keep the #31 open

Issue #34, opened 2019-12-19
WARNING: ThreadSanitizer: data race


ChunMinChang:
Running `RUSTFLAGS="-Z sanitizer=thread" cargo test` or the executable it generates will shows the following data-race warning

```
(lldb) r
Process 8952 launched: '/Users/cchang/Work/cubeb/src/cubeb-coreaudio-rs/target/debug/deps/cubeb_coreaudio-c4da6e229b8adc8d' (x86_64)

running 190 tests
test backend::auto_array::test_auto_array ... ok
test backend::tests::aggregate_device::test_aggregate_activate_clock_drift_compensation ... ignored
test backend::tests::aggregate_device::test_aggregate_activate_clock_drift_compensation_for_a_blank_aggregate_device ... ignored
test backend::tests::aggregate_device::test_aggregate_activate_clock_drift_compensation_for_an_aggregate_device_without_master_device ... ignored
test backend::auto_release::test_auto_release ... ok
test backend::tests::aggregate_device::test_aggregate_create_blank_device ... ignored
test backend::tests::aggregate_device::test_aggregate_activate_clock_drift_compensation_for_an_unknown_aggregate_device ... ok
test backend::tests::aggregate_device::test_aggregate_destroy_aggregate_device_for_a_unknown_plugin_device ... ignored
test backend::tests::aggregate_device::test_aggregate_destroy_device_for_unknown_plugin_and_aggregate_devices ... ok
test backend::tests::aggregate_device::test_aggregate_get_sub_devices_for_blank_aggregate_devices ... ignored
test backend::tests::aggregate_device::test_aggregate_set_master_device ... ignored
test backend::tests::aggregate_device::test_aggregate_set_master_device_for_a_blank_aggregate_device ... ignored
test backend::tests::aggregate_device::test_aggregate_get_sub_devices_for_a_unknown_device ... ok
test backend::tests::aggregate_device::test_aggregate_set_sub_devices ... ignored
test backend::tests::aggregate_device::test_aggregate_set_master_device_for_an_unknown_aggregate_device ... ok
test backend::tests::aggregate_device::test_aggregate_set_sub_devices_for_unknown_input_devices ... ignored
test backend::tests::aggregate_device::test_aggregate_set_sub_devices_for_unknown_output_devices ... ignored
test backend::tests::aggregate_device::test_aggregate_set_sub_devices_for_unknown_devices ... ok
[...]

PR #46, opened 2020-02-10
Run tests by sanitizers

ChunMinChang:
- Add a script to run test suits by all the available sanitizers
- Add an instruction of running sanitizers in _README.md_
- Run the sanitizers on _Travis CI_ but the tests are allowed to fail
  - Since sanitizers are under development, they may false positives, or be unavailable for Nightly toolchain sometimes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,8 @@ before_script:
 script:
   - cargo build --verbose
   - sh run_tests.sh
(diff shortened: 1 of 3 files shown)


ChunMinChang:
This will be helpers for #34 and #45

padenot (approved):

PR #47, opened 2020-02-10
Replace`*.get_mut()` by `*.store()` on all the `Atomic*`

ChunMinChang:
`Atomic*::get_mut()` uses `UnsafeCell` to mutate the value directly, so `Atomic*::get_mut()` is only safe if it's called from a single thread. If `Atomic*::get_mut()` is called from multiple threads, or `Atomic*` is accessible as mutable references in different threads, it will cause the data race. 
Thus, for all the `Atomic*` in the current code, the `*.get_mut()` should be replaced by `*.store()` to avoid the data race. This change will solve the data race found by ThreadSanitizer(#34).

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -794,7 +794,7 @@ extern "C" fn audiounit_property_listener_callback(
         );
         return NO_ERR;
     }
-    *stm.switching_device.get_mut() = true;
+    stm.switching_device.store(true, Ordering::SeqCst);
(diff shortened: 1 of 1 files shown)


ChunMinChang:
@kinetiknz : Could you review this ?

kinetiknz (approved):
LGTM, thanks!

padenot (approved):

PR #49, opened 2020-02-11
Clean up clippy warnings and errors

### LINKED ISSUE #43: Fix the clippy warnings
We have lots of _clippy_ warnings that need to be fixed. We should add the _clippy_ check in the *.travis.yml* as what we do in [audio-mixer](https://github.com/ChunMinChang/audio-mixer/blob/956990624e7ce4144692525d311d7f82632b30a3/.travis.yml).

ChunMinChang:
Huge clean up for the *clippy* warnings and errors and enable the `cargo fmt` and `cargo clippy` checks on Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,7 @@ env:
 before_script:
   - rustc --version
   - cargo --version
-  - rustup component add rustfmt
-  - rustup component add clippy
+  - sh install_rustfmt_clippy.sh
 script:
   - cargo fmt --all -- --check
   - cargo clippy -- -D warnings
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ rust:
   - nightly
 os:
   - osx
+env:
+  - RUST_BACKTRACE=1
 before_script:
   - rustc --version
   - cargo --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ env:
 before_script:
   - rustc --version
   - cargo --version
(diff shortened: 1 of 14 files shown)


ChunMinChang:
By
- https://rust-lang-nursery.github.io/rust-toolstate/
- https://rust-lang.github.io/rustup-components-history/x86_64-apple-darwin.html

*rustfmt* is not available on Rust nightly today on *Mac OSX*

ChunMinChang:
This should address #43

.travis.yml
```diff
@@ -5,9 +5,15 @@ rust:
   - nightly
 os:
   - osx
+env:
+  - RUST_BACKTRACE=1
 before_script:
   - rustc --version
   - cargo --version
+  - rustup component add rustfmt
+  - rustup component add clippy
```
padenot: we should investigate if we should pin specific versions of this to have a setup that keeps running when tools change.

padenot (approved):

.travis.yml
ChunMinChang: I add a [script](https://github.com/ChunMinChang/cubeb-coreaudio-rs/commit/3a76d60e8886281e5329ba4f3bd3c64c36d49dff#diff-376b3da8cd65e00f28182b7056b62f8a) to install the latest available *clippy*/*rustfmt*.

The rust nightly will be installed to the latest version that *clippy* works. And the *rustfmt* will also be installed to that version, since installing *clippy* requires *rustfmt*.

PR #50, opened 2020-02-12
Bail out if the output device has no output channel

ChunMinChang:
We should return an error when the output hardware has no output channel. Then we won't create a zero-output audio mixer and hit the assertion. It solves the [BMO 1613156](https://bugzilla.mozilla.org/show_bug.cgi?id=1613156).


padenot (approved):
Hopefully we'll figure out what's going on one day.

Issue #28, opened 2019-11-27
Boost the speed when opening multiple streams simultaneously


ChunMinChang:
In the current _C_ implementation, each time when the stream starts, we will first [try locking the mutex within cubeb context][c] to avoid racing with `reinit`/`destroy`/`stop`, then call stream-start.

In the current _Rust_ implementation, each time when the stream starts, we will [dispatch this operation to the same task queue][rust] where `reinit`/`destroy`/`stop` is on to avoid racing issues.

Starting a cubeb stream in both implementations above requires sharing a resource in its cubeb context to avoid racing issues. For this reason, when opening multiple streams (in same cubeb context) at the same time, the stream-start will be executed sequentially, while they could actually be run in parallel since they are not sharing data.

To boost the speed when opening multiple streams at the same time, we could
1. either move the task queue for stream operations, [`serial_queue`][queue], from cubeb context to cubeb stream
    - This might relate to the refactoring plan: [separate the audio stream into inner and outer][separate-stream] 
2. or use a dummy mutex, `Mutex<()>`, as the token to operate [`core_stream_data`][core_stream_data] to avoid [racing issues][racing]
    - Ideally we should just change [`core_stream_data`][core_stream_data] to `Mutex<CoreStreamData>`, but it's likely to cause [deadlocks][deadlock]
    - workaround for above deadlock: Using `Mutex<()>` for `CoreStreamData`
        - except callback thread, [`core_stream_data`][core_stream_data] should only be changed within the thread that locks the a dummy mutex: `Mutex<()>`
        - to avoid racing with callback thread, just stop the _AudioUnit_ in the right place. Stopping  _AudioUnit_ will acquire the same mutex inside _AudioUnit_ that is used when it's data callback is run.

Both approaches can be implemented easily in the current _Rust_ code. From my test, they give similar speeds. They are both around _6x_ faster than the current _Rust_ implementation, when opening and starting _10_ streams at the same time. For opening one stream at a time, the above approaches doesn't bring much benefit. 

This problem is not urgent since (I believe) users don't often open multiple streams at the same time. We could fix this when this becomes a real issue. This is more like a note for now.

[racing]: https://bugzilla.mozilla.org/show_bug.cgi?id=1598413#c1

[deadlock]: https://github.com/ChunMinChang/cubeb-coreaudio-rs/pull/26#discussion_r350886537

[...]

PR #55, opened 2020-02-28
Rework threading model

ChunMinChang:
The change made here can address #28, and make sure that no stream task can be executed after the stream is destroyed, which should avoid the potential data race.

The patch creates a struct `Queue` wrapping `dispatch_queue_t`, which will retain and release the reference count of `dispatch_queue_t` when it's created/cloned, or destroyed. By creating the task queue in each stream, the stream's task that don't need to share the common data among all streams can be executed in each stream's own task queue. This change also brings another benefit: now we can know what task can be canceled. All the tasks appended after the stream-destroy task should be canceled.

### DIFF
--- a/coreaudio-sys-utils/src/dispatch.rs
+++ b/coreaudio-sys-utils/src/dispatch.rs
@@ -68,9 +68,14 @@ impl Queue {
         set_dispatch_context(self.0, context);
     }
 
-    // This will release the inner `dispatch_queue_t` asynchronously.
     fn release(&self) {
-        release_dispatch_queue(self.0);
+        unsafe {
+            // This will release the inner `dispatch_queue_t` asynchronously.
+            // TODO: This is incredibly unsafe. Find another way to release the queue.
+            dispatch_release(mem::transmute::<dispatch_queue_t, dispatch_object_t>(
+                self.0,
+            ));
+        }
     }
 }
 
@@ -89,13 +94,6 @@ impl Clone for Queue {
 
 // Low-level Grand Central Dispatch (GCD) APIs
 // ------------------------------------------------------------------------------------------------
-fn release_dispatch_queue(queue: dispatch_queue_t) {
-    // TODO: This is incredibly unsafe. Find another way to release the queue.
-    unsafe {
-        dispatch_release(mem::transmute::<dispatch_queue_t, dispatch_object_t>(queue));
-    }
-}
-
 fn retain_dispatch_queue(queue: dispatch_queue_t) {
     // TODO: This is incredibly unsafe. Find another way to retain the queue.
     unsafe {
--- a/coreaudio-sys-utils/src/dispatch.rs
+++ b/coreaudio-sys-utils/src/dispatch.rs
@@ -5,6 +5,51 @@ use std::mem;
 use std::os::raw::c_void;
 use std::ptr;
 
+// Queue: A wrapper around `dispatch_queue_t`.
+// ------------------------------------------------------------------------------------------------
+#[derive(Debug)]
+pub struct Queue(dispatch_queue_t);
+
+impl Queue {
+    pub fn new(label: &'static str) -> Self {
+        Self(create_dispatch_queue(label, DISPATCH_QUEUE_SERIAL))
+    }
+
+    pub fn run_async<F>(&self, work: F)
+    where
+        F: Send + FnOnce(),
+    {
+        async_dispatch(self.0, work);
+    }
+
+    pub fn run_sync<F>(&self, work: F)
+    where
+        F: Send + FnOnce(),
+    {
+        sync_dispatch(self.0, work);
+    }
+
+    // This will release the inner `dispatch_queue_t` asynchronously.
+    fn release(&self) {
+        release_dispatch_queue(self.0);
+    }
+}
+
+impl Drop for Queue {
+    fn drop(&mut self) {
+        self.release();
+    }
+}
+
+impl Clone for Queue {
+    fn clone(&self) -> Self {
+        retain_dispatch_queue(self.0);
+        Self(self.0)
+    }
+}
+
+// Low-level Grand Central Dispatch (GCD) APIs
+// ------------------------------------------------------------------------------------------------
 pub const DISPATCH_QUEUE_SERIAL: dispatch_queue_attr_t = ptr::null_mut::<dispatch_queue_attr_s>();
 
 pub fn create_dispatch_queue(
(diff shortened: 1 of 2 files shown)


ChunMinChang:
result on tryserver: https://treeherder.mozilla.org/#/jobs?repo=try&revision=f68714a4cf88731e50fc9278d0d6ad78647bf98c&selectedJob=291020567

-------------------
The error above is intermittent I think. It's robust in the today's tryserver: https://treeherder.mozilla.org/#/jobs?repo=try&revision=9d125519db4836e37b0c069c41ab4f79c3c01129

coreaudio-sys-utils/src/dispatch.rs
```diff
@@ -4,26 +4,110 @@ use std::ffi::CString;
 use std::mem;
 use std::os::raw::c_void;
 use std::ptr;
+use std::sync::atomic::{AtomicBool, Ordering};
 
-pub const DISPATCH_QUEUE_SERIAL: dispatch_queue_attr_t = ptr::null_mut::<dispatch_queue_attr_s>();
+// Queue: A wrapper around `dispatch_queue_t`.
+// ------------------------------------------------------------------------------------------------
+#[derive(Debug)]
```
padenot, line 11: Do we need the `derive(Debug)` here? I seem to recall there was an issue with code size.

coreaudio-sys-utils/src/dispatch.rs
```diff
@@ -4,26 +4,110 @@ use std::ffi::CString;
...
+        retain_dispatch_queue(self.0);
+        Self(self.0)
+    }
+}
+
+// Low-level Grand Central Dispatch (GCD) APIs
+// ------------------------------------------------------------------------------------------------
+const DISPATCH_QUEUE_SERIAL: dispatch_queue_attr_t = ptr::null_mut::<dispatch_queue_attr_s>();
+
+fn create_dispatch_queue(label: &str, queue_attr: dispatch_queue_attr_t) -> dispatch_queue_t {
```
padenot: Have you thought about inlining all of those in the queue itself?

coreaudio-sys-utils/src/dispatch.rs
```diff
@@ -4,26 +4,110 @@ use std::ffi::CString;
...
     // TODO: This is incredibly unsafe. Find another way to release the queue.
     unsafe {
         dispatch_release(mem::transmute::<dispatch_queue_t, dispatch_object_t>(queue));
     }
 }
 
-pub fn async_dispatch<F>(queue: dispatch_queue_t, work: F)
+fn retain_dispatch_queue(queue: dispatch_queue_t) {
+    // TODO: This is incredibly unsafe. Find another way to retain the queue.
+    unsafe {
```
padenot: It would be less unsafe if you had everything in the same class.

coreaudio-sys-utils/src/dispatch.rs
```diff
@@ -4,26 +4,110 @@ use std::ffi::CString;
...
+const DISPATCH_QUEUE_SERIAL: dispatch_queue_attr_t = ptr::null_mut::<dispatch_queue_attr_s>();
+
+fn create_dispatch_queue(label: &str, queue_attr: dispatch_queue_attr_t) -> dispatch_queue_t {
     let label = CString::new(label).unwrap();
     let c_string = label.as_ptr();
     unsafe { dispatch_queue_create(c_string, queue_attr) }
 }
 
-pub fn release_dispatch_queue(queue: dispatch_queue_t) {
+fn release_dispatch_queue(queue: dispatch_queue_t) {
```
padenot: Would be best to have a refcount class here with a drop, etc., to ensure release and retain are balanced.

coreaudio-sys-utils/src/dispatch.rs
```diff
@@ -43,6 +127,32 @@ where
...
+fn get_dispatch_context<'a, T>(queue: dispatch_queue_t) -> Option<&'a mut T> {
+    unsafe {
+        let context =
+            dispatch_get_context(mem::transmute::<dispatch_queue_t, dispatch_object_t>(queue))
+                as *mut T;
+        context.as_mut()
+    }
+}
+
+fn set_dispatch_context<T>(queue: dispatch_queue_t, context: Box<T>) {
```
padenot: I don't see the purpose of having two layers here really (functions + struct). Wouldn't it be better with everything inlined, and no premature generalization? You only need this for bool.

coreaudio-sys-utils/src/dispatch.rs
```diff
@@ -4,26 +4,110 @@ use std::ffi::CString;
...
+            }
+            work();
+        });
+    }
+
+    pub fn run_final<F>(&self, work: F)
+    where
+        F: Send + FnOnce(),
+    {
+        self.set_context(Box::new(AtomicBool::new(false)));
```
padenot: This should be done in the ctor, right? Then you can remove the API for context management.

padenot (changes requested):
We need this less generic, we don't need it.

coreaudio-sys-utils/src/dispatch.rs
ChunMinChang: Yes. `#[derive(Debug)]` is applied in `AudioUnitStream` and `AudioUnitContext` as well. As `Queue` is a member of those _strcut_ implementing `Debug` trait, `Queue` need to implement the `Debug` trait as well. Otherwise, there will be an error like:
```
`Queue` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`
```

coreaudio-sys-utils/src/dispatch.rs
ChunMinChang: The main purpose to split them is to make them look clearer. I can merge those functions into the struct and see how it goes

ChunMinChang:
Try results with new patches: https://treeherder.mozilla.org/#/jobs?repo=try&selectedJob=293110963&revision=a1e4f7ac0f6103fe2e1dab2b4fd1522814952b2e

coreaudio-sys-utils/src/dispatch.rs
padenot: Maybe it's best to remove those as well. I thought we had a policy about that, I think on dev-platform?

coreaudio-sys-utils/src/dispatch.rs
```diff
@@ -4,148 +4,201 @@ use std::ffi::CString;
...
+impl Drop for Queue {
+    fn drop(&mut self) {
+        self.release();
+    }
+}
 
-        callback(queue, resource);
+impl Clone for Queue {
+    fn clone(&self) -> Self {
+        // TODO: This is incredibly unsafe. Find another way to retain the queue.
```
padenot: Can you write why it is unsafe in this comment so it's clearer what you mean?

padenot (approved):

coreaudio-sys-utils/src/dispatch.rs
ChunMinChang: I'll open a followup PR to remove `#[derive(Debug)]` to avoid adding debug implementation stuff into the Firefox release version.

coreaudio-sys-utils/src/dispatch.rs
ChunMinChang: The comment is around calling `transmute` directly. I'll revise the comments.

PR #57, opened 2020-03-23
Fix two bugs and add logs

padenot:
This is about duplex stream that have to use non-aggregate devices.

Bug #1 (https://github.com/ChunMinChang/cubeb-coreaudio-rs/commit/89147a72e90716d3e4cd37230d4d8f166c15c506) was that we're not draining the `linear_input_buffer`. If the output is super long to start compared to the output, a lot of input audio has been buffered. We need to remove most of the input frames when we have the first output callback. The opposite case was covered, not having audio input data when the first output callback was early.

Bug #2 (https://github.com/ChunMinChang/cubeb-coreaudio-rs/commit/5499668a61454dc7f51bd78ad13ab94b8c1d3f0e) was that the code was completely wrong, the function is to be able to feed the right amount of data into the resampler, to have continuous audio, and to detect when we're under-running, to try to correct it. This is the right code. The C++ has the same bug so it's not cause by rewrite I guess.

Then I'm adding some logs that I found useful to debug (https://github.com/ChunMinChang/cubeb-coreaudio-rs/commit/5499668a61454dc7f51bd78ad13ab94b8c1d3f0e).

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -642,6 +642,8 @@ extern "C" fn audiounit_output_callback(
             }
         };
 
+        let prev_frames_written = stm.frames_written.load(Ordering::SeqCst);
+
         stm.frames_written
             .fetch_add(i64::from(output_frames), Ordering::SeqCst);
 
@@ -661,6 +663,26 @@ extern "C" fn audiounit_output_callback(
                 frames_written,
             );
             let missing_frames = input_frames_needed - stm.frames_read.load(Ordering::SeqCst);
+            // Else if the input has buffered a lot already because the output started late, we
+            // need to trim the input buffer
+            let input_channels = stm.core_stream_data.input_desc.mChannelsPerFrame as usize;
+            let buffered_input_frames = stm
+                .core_stream_data
+                .input_linear_buffer
+                .as_ref()
+                .unwrap()
+                .elements()
+                / input_channels;
+            if prev_frames_written == 0 && buffered_input_frames > input_frames_needed  as usize {
+                let samples_to_pop = (buffered_input_frames - input_frames_needed as usize) * input_channels;
+                stm.core_stream_data
+                    .input_linear_buffer
+                    .as_mut()
+                    .unwrap()
+                    .pop(samples_to_pop);
+                stm.frames_read.fetch_sub((samples_to_pop / input_channels) as i64, Ordering::SeqCst);
+            }
+
             let elements = (missing_frames
                 * i64::from(stm.core_stream_data.input_desc.mChannelsPerFrame))
                 as usize;
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -660,7 +660,7 @@ extern "C" fn audiounit_output_callback(
             let input_frames_needed = minimum_resampling_input_frames(
                 stm.core_stream_data.input_hw_rate,
                 f64::from(stm.core_stream_data.output_stream_params.rate()),
-                frames_written,
+                output_frames as i64,
             );
             let missing_frames = input_frames_needed - stm.frames_read.load(Ordering::SeqCst);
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -204,15 +204,18 @@ fn create_device_info(id: AudioDeviceID, devtype: DeviceType) -> Result<device_i
 
     let default_device_id = audiounit_get_default_device_id(devtype);
     if default_device_id == kAudioObjectUnknown {
+        cubeb_log!("Could not find default audio device for {:?}", devtype);
         return Err(Error::error());
     }
 
     if id == kAudioObjectUnknown {
         info.id = default_device_id;
+        cubeb_log!("Falling back to default device, after having requested a specific device.");
```
ChunMinChang, line 213: I guess `"after having requested a specific device."` isn't always true. The users can set a `nullptr` `DeviceId`, which is a `cubeb_devid` = `*const c_void`, as default value to request using default device.

src/backend/mod.rs
```diff
@@ -204,15 +204,18 @@ fn create_device_info(id: AudioDeviceID, devtype: DeviceType) -> Result<device_i
...
         info.flags |= device_flags::DEV_SELECTED_DEFAULT;
     }
 
     if info.id == default_device_id {
+        cubeb_log!("Requesting default input device.");
```
ChunMinChang, line 218: This can not only happen for an input device. `create_device_info` can be called for both input and output. It should be `cubeb_log!("Requesting default {:?} device", devtype)`.

src/backend/mod.rs
```diff
@@ -669,7 +672,7 @@ extern "C" fn audiounit_output_callback(
                     .push_zeros(elements);
                 stm.frames_read.store(input_frames_needed, Ordering::SeqCst);
                 cubeb_log!(
-                    "({:p}) {} pushed {} frames of input silence.",
+                    "({:p}) Missing Frames {} pushed {} frames of input silence.",
```
ChunMinChang, line 697: `"({:p}) Missing Frames: {} pushed {} frames of input silence."` might be better. 

Otherwise, the log can be "Missing Frames Input hasn't started, ...".

src/backend/mod.rs
```diff
@@ -661,6 +663,26 @@ extern "C" fn audiounit_output_callback(
...
+            // need to trim the input buffer
+            let input_channels = stm.core_stream_data.input_desc.mChannelsPerFrame as usize;
+            let buffered_input_frames = stm
+                .core_stream_data
+                .input_linear_buffer
+                .as_ref()
+                .unwrap()
+                .elements()
+                / input_channels;
+            if prev_frames_written == 0 && buffered_input_frames > input_frames_needed  as usize {
```
ChunMinChang, line 676: Maybe we should make `minimum_resampling_input_frames` return an `usize` instead of using `as` everywhere?

src/backend/mod.rs
```diff
@@ -655,9 +660,29 @@ extern "C" fn audiounit_output_callback(
             let input_frames_needed = minimum_resampling_input_frames(
                 stm.core_stream_data.input_hw_rate,
                 f64::from(stm.core_stream_data.output_stream_params.rate()),
-                frames_written,
+                output_frames as i64,
```
ChunMinChang, line 663: `let frames_written = stm.frames_written.load(Ordering::SeqCst);` now can be removed, right?

ChunMinChang:
I am ok to merge this first and import it to Firefox ASAP and fix the comments later. Unfortunately, I don't have level-3 to push the commit. I'll see if I can find one to vouch for me. If I can find one today, I'll merge this first.

ChunMinChang:
Merge the update first so users can test if this fix their issues. We should file another PR as a follow up.

ChunMinChang:
I think it's better to fix #61 before importing this to Firefox.

PR #63, opened 2020-03-24
Fix comments in #57

padenot:
This is based on #62  and fixes the comments in #57.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -210,12 +210,12 @@ fn create_device_info(id: AudioDeviceID, devtype: DeviceType) -> Result<device_i
 
     if id == kAudioObjectUnknown {
         info.id = default_device_id;
-        cubeb_log!("Falling back to default device, after having requested a specific device.");
+        cubeb_log!("Creating a default device info.");
         info.flags |= device_flags::DEV_SELECTED_DEFAULT;
     }
 
     if info.id == default_device_id {
-        cubeb_log!("Requesting default input device.");
+        cubeb_log!("Requesting default system device.");
         info.flags |= device_flags::DEV_SYSTEM_DEFAULT;
     }
 
@@ -694,14 +694,14 @@ extern "C" fn audiounit_output_callback(
                     .push_zeros(silent_samples_to_push);
                 stm.frames_read.fetch_add(input_frames_needed, Ordering::SeqCst);
                 cubeb_log!(
-                    "({:p}) Missing Frames {} pushed {} frames of input silence.",
+                    "({:p}) Missing Frames: {} pushed {} frames of input silence.",
                     stm.core_stream_data.stm_ptr,
                     if stm.frames_read.load(Ordering::SeqCst) == 0 {
-                        "Input hasn't started,"
+                        "input hasn't started,"
                     } else if stm.switching_device.load(Ordering::SeqCst) {
-                        "Device switching,"
+                        "device switching,"
                     } else {
-                        "Drop out,"
+                        "drop out,"
                     },
                     silent_frames_to_push
                 );
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -42,7 +42,7 @@ use std::mem;
 use std::os::raw::c_void;
 use std::ptr;
 use std::slice;
-use std::sync::atomic::{AtomicBool, AtomicUsize, AtomicU32, AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU32, AtomicU64, AtomicUsize, Ordering};
 use std::sync::{Arc, Condvar, Mutex};
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


src/backend/mod.rs
```diff
@@ -645,65 +645,65 @@ extern "C" fn audiounit_output_callback(
...
-            if missing_frames > 0 {
+            if missing_input_frames {
+                let silent_frames_to_push = input_frames_needed - buffered_input_frames;
+                let silent_samples_to_push = silent_frames_to_push / input_channels;
                 stm.core_stream_data
                     .input_linear_buffer
                     .as_mut()
                     .unwrap()
-                    .push_zeros(elements);
-                stm.frames_read.store(input_frames_needed, Ordering::SeqCst);
```
padenot, line 696: This was a bug, we need to add here, since we're pushing frames to the stream.

src/backend/mod.rs
```diff
@@ -659,19 +660,18 @@ extern "C" fn audiounit_output_callback(
...
                 .as_ref()
                 .unwrap()
                 .elements()
                 / input_channels;
+            let missing_input_frames = input_frames_needed > buffered_input_frames;
+            // Else if the input has buffered a lot already because the output started late, we
+            // need to trim the input buffer
             if prev_frames_written == 0 && buffered_input_frames > input_frames_needed as usize {
                 let samples_to_pop =
                     (buffered_input_frames - input_frames_needed as usize) * input_channels;
```
ChunMinChang, line 680: Does `buffered_input_frames` need to be updated, after `input_linear_buffer` is popped some samples out? or using some scope to limit the lifetime of `buffered_input_frames`? or adding a comment mentioning `buffered_input_frames` cannot be used anymore at least?

If we have a struct wrapping all the frame, sample, channels,.. stuff, so we don't need to worry about synchronizing the `frames_read`, `input_linear_buffer`, ... etc. We can do this in a follow-up or just mention it in the todo list.

src/backend/mod.rs
```diff
@@ -659,19 +660,18 @@ extern "C" fn audiounit_output_callback(
...
-            // need to trim the input buffer
-            let input_channels = stm.core_stream_data.input_desc.mChannelsPerFrame as usize;
             let buffered_input_frames = stm
                 .core_stream_data
                 .input_linear_buffer
                 .as_ref()
                 .unwrap()
                 .elements()
                 / input_channels;
+            let missing_input_frames = input_frames_needed > buffered_input_frames;
```
ChunMinChang: I'd prefer put this above `if missing_input_frames {` since `missing_input_frames` doesn't be used in between.

src/backend/mod.rs
```diff
@@ -681,19 +681,18 @@ extern "C" fn audiounit_output_callback(
...
+                    .fetch_sub((samples_to_pop / input_channels) as usize, Ordering::SeqCst);
             }
 
-            let elements = (missing_frames
-                * i64::from(stm.core_stream_data.input_desc.mChannelsPerFrame))
-                as usize;
-            if missing_frames > 0 {
+            if missing_input_frames {
+                let silent_frames_to_push = input_frames_needed - buffered_input_frames;
+                let silent_samples_to_push = silent_frames_to_push / input_channels;
```
ChunMinChang: Why `silent_samples_to_push` is `silent_frames_to_push / input_channels` instead of `silent_frames_to_push * input_channels`?

src/backend/mod.rs
ChunMinChang: Good catch! Thanks!

ChunMinChang (changes requested):

src/backend/mod.rs
padenot: Good catch, it's a bug. Tests didn't catch it because my macbook has its microphone set to Mono :-(. I'm changing that.

padenot: Hrm weird my Macbook Pro 15" 2018 only has a mono microphone, but my Macbook Pro 15" 2016 has the option of doing mono, stereo, or even quad!

src/backend/mod.rs
padenot: > Does `buffered_input_frames` need to be updated, after `input_linear_buffer` is popped some samples out? or using some scope to limit the lifetime of `buffered_input_frames`? or adding a comment mentioning `buffered_input_frames` cannot be used anymore at least?

It doesn't matter: if we're popping frames from the linear input queue, then we have more than enough input frames for this callback (by definition, otherwise we wouldn't be doing that). If this is the case, then we're never be in the case where we're missing frames (because we just set the linear input buffer to be the exact right size).

src/backend/mod.rs
```diff
@@ -684,9 +683,9 @@ extern "C" fn audiounit_output_callback(
                     .fetch_sub((samples_to_pop / input_channels) as usize, Ordering::SeqCst);
             }
 
-            if missing_input_frames {
+            if input_frames_needed > buffered_input_frames;
```
ChunMinChang: Oh `;` is a typo. It should be `{`:
`if input_frames_needed > buffered_input_frames {`

ChunMinChang (approved):

PR #64, opened 2020-03-24
Try using the model id as group id.

padenot:
It works for the all the external devices that I have, but not for the
built-in mic/speakers and the trrs plug of the macbook. I wonder if we could hardcode those.

### DIFF
--- a/src/backend/device_property.rs
+++ b/src/backend/device_property.rs
@@ -21,6 +21,23 @@ pub fn get_device_uid(
     }
 }
 
+pub fn get_device_model_uid(
+    id: AudioDeviceID,
+    devtype: DeviceType,
+) -> std::result::Result<StringRef, OSStatus> {
+    assert_ne!(id, kAudioObjectUnknown);
+
+    let address = get_property_address(Property::ModelUID, devtype);
+    let mut size = mem::size_of::<CFStringRef>();
+    let mut uid: CFStringRef = ptr::null();
+    let err = audio_object_get_property_data(id, &address, &mut size, &mut uid);
+    if err == NO_ERR {
+        Ok(StringRef::new(uid as _))
+    } else {
+        Err(err)
+    }
+}
+
 pub fn get_device_source(
     id: AudioDeviceID,
     devtype: DeviceType,
@@ -276,6 +293,7 @@ pub enum Property {
     DeviceStreamFormat,
     DeviceStreams,
     DeviceUID,
+    ModelUID,
     HardwareDefaultInputDevice,
     HardwareDefaultOutputDevice,
     HardwareDevices,
@@ -298,6 +316,7 @@ impl From<Property> for AudioObjectPropertySelector {
             Property::DeviceStreamFormat => kAudioDevicePropertyStreamFormat,
             Property::DeviceStreams => kAudioDevicePropertyStreams,
             Property::DeviceUID => kAudioDevicePropertyDeviceUID,
+            Property::ModelUID => kAudioDevicePropertyModelUID,
             Property::HardwareDefaultInputDevice => kAudioHardwarePropertyDefaultInputDevice,
             Property::HardwareDefaultOutputDevice => kAudioHardwarePropertyDefaultOutputDevice,
             Property::HardwareDevices => kAudioHardwarePropertyDevices,
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -1506,11 +1506,25 @@ fn create_cubeb_device_info(
         Ok(uid) => {
             let c_string = uid.into_cstring();
             dev_info.device_id = c_string.into_raw();
-            dev_info.group_id = dev_info.device_id;
         }
         Err(e) => {
             cubeb_log!(
                 "Cannot get the uid for device {} in {:?} scope. Error: {}",
+                devid,
+                devtype,
+                e
+            );
+        }
+    }
+
+    match get_device_model_uid(devid, devtype) {
+        Ok(uid) => {
+            let c_string = uid.into_cstring();
+            dev_info.group_id = c_string.into_raw();
+        }
+        Err(e) => {
+            cubeb_log!(
+                "Cannot get the model uid for device {} in {:?} scope. Error: {}",
                 devid,
                 devtype,
                 e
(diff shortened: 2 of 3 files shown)


src/backend/device_property.rs
```diff
@@ -21,6 +21,23 @@ pub fn get_device_uid(
     }
 }
 
+pub fn get_device_model_uid(
+    id: AudioDeviceID,
+    devtype: DeviceType,
+) -> std::result::Result<StringRef, OSStatus> {
+    assert_ne!(id, kAudioObjectUnknown);
+
+    let address = get_property_address(Property::DeviceManufacturer, devtype);
```
ChunMinChang: `get_property_address(Property::ModelUID, devtype)`

ChunMinChang (commented):
LGTM, but I think this won't pass the test since we have a check for `info.group_id == info.device_id`

https://github.com/ChunMinChang/cubeb-coreaudio-rs/blob/3860bb3f957b3be1d56a01f9e73d5d32e463eda0/src/backend/tests/api.rs#L1397

ChunMinChang:
And we need to retake the leaked `dev_info.group_id` in `destroy_cubeb_device_info`

https://github.com/ChunMinChang/cubeb-coreaudio-rs/blob/3860bb3f957b3be1d56a01f9e73d5d32e463eda0/src/backend/mod.rs#L1636

------------------

No, I was wrong.

src/backend/mod.rs
```diff
@@ -1631,10 +1645,8 @@ fn destroy_cubeb_device_info(device: &mut ffi::cubeb_device_info) {
     // Set the pointers to null in case it points to some released memory.
     unsafe {
         if !device.device_id.is_null() {
-            // group_id is a mirror to device_id, so we could skip it.
-            assert!(!device.group_id.is_null());
-            assert_eq!(device.device_id, device.group_id);
             let _ = CString::from_raw(device.device_id as *mut _);
+            let _ = CString::from_raw(device.group_id as *mut _);
```
ChunMinChang: It's better to change it to
```rust
if !device.group_id.is_null() {
    let _ = CString::from_raw(device.group_id as *mut _);
    device.group_id = ptr::null();
}
```

padenot: Is that only needed for the group id you think ?

ChunMinChang: And we need to change here as well:

https://github.com/ChunMinChang/cubeb-coreaudio-rs/blob/3860bb3f957b3be1d56a01f9e73d5d32e463eda0/src/backend/tests/api.rs#L1497-L1499

Otherwise the `device.device_id` will be freed twice by 
```rust
let _ = CString::from_raw(device.device_id as *mut _);
```
and then
```rust
let _ = CString::from_raw(device.group_id as *mut _);
```

padenot: But they are different strings now?

padenot: ah I see this is the test

ChunMinChang: They will be the same pointer in `test_device_destroy`

padenot: Best to fix the test for new invariants.

ChunMinChang: > Best to fix the test for new invariants.

Yes

On the other hand, we can remove `test_device_destroy_with_different_device_id_and_group_id` as well. It's used to check the `group_id` is the same as `device_id`. We don't need it now.

src/backend/mod.rs
```diff
@@ -1631,12 +1645,12 @@ fn destroy_cubeb_device_info(device: &mut ffi::cubeb_device_info) {
...
-            // group_id is a mirror to device_id, so we could skip it.
-            assert!(!device.group_id.is_null());
-            assert_eq!(device.device_id, device.group_id);
             let _ = CString::from_raw(device.device_id as *mut _);
             device.device_id = ptr::null();
-            device.group_id = ptr::null();
+            if !device.group_id.is_null() {
+                let _ = CString::from_raw(device.group_id as *mut _);
+                device.group_id = ptr::null();
+            }
```
```rust
if !device.group_id.is_null() {
    let _ = CString::from_raw(device.group_id as *mut _);
    device.group_id = ptr::null();
}
```
should be put below
```rust
if !device.device_id.is_null() {
    ...
}
```
since `group_id` isn't relied on `device_id`anymore.

src/backend/tests/api.rs
```diff
@@ -1394,7 +1394,6 @@ fn test_create_cubeb_device_info() {
         assert_eq!(info.devid as AudioObjectID, id);
         assert!(!info.device_id.is_null());
         assert!(!info.friendly_name.is_null());
-        assert_eq!(info.group_id, info.device_id);
```
ChunMinChang, line 1397: It's better to call `assert!(!info.group_id.is_null())`. 

`check_device_info_by_device` will only be called if `create_cubeb_device_info` returns `Ok(ffi::cubeb_device_info)` so `group_id` must be pointed to some leaked `CString` (by `into_raw`).

ChunMinChang (changes requested):
The test fails on my end (and Travis server [here](https://travis-ci.org/github/ChunMinChang/cubeb-coreaudio-rs/jobs/666508423#L396)). The following things should be fixed:

1.`test_device_destroy_with_different_device_id_and_group_id` should be removed 
2. in `destroy_cubeb_device_info`, the `group_id` should be assigned to something like

```rust
let group_id = CString::new("test:group id").unwrap();
...
device.group_id = group_id.into_raw();
```

src/backend/mod.rs
padenot: Yes, but is it useful to allow null pointers here? Can't we enforce that the pointer is always valid (maybe to an empty string?).

padenot: We seem to enforce it, in the test, so I don't know why we need this check. Maybe an assert would be better to document the invariant.

src/backend/tests/api.rs
```diff
@@ -1496,7 +1497,7 @@ fn test_device_destroy() {
 
     device.device_id = device_id.into_raw();
     // The group_id is a mirror to device_id in our implementation, so we could skip it.
-    device.group_id = device.device_id;
+    device.group_id = device.device_id.clone();
```
ChunMinChang: If the `device.device_id` is `0xfeedfade`, then `device.group_id` is also `0xfeedfade`. Then both
`let _ = CString::from_raw(device.device_id as *mut _)` and `let _ = CString::from_raw(device.group_id as *mut _)` will be called. As a result, the `0xfeedfade` will be released twice.

I guess something like below could work:
```rust
let group_id = CString::new("test: group id").unwrap();
device.group_id = group_id.into_raw();
```

src/backend/mod.rs
ChunMinChang: Agree. Could you replace those `if` to `assert`?

ChunMinChang (approved):

PR #66, opened 2020-03-25
Add a hook to check that commits are formatted and linted correctly

padenot:
Have a look, see if you like it. I've left the doc in the file `pre-push`.

Users need to run `install_git_hook.sh` to install this hook.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,42 @@
+#!/bin/sh -e
+
+# An example hook script to verify what is about to be pushed.  Called by "git
+# push" after it has checked the remote status, but before anything has been
+# pushed.  If this script exits with a non-zero status nothing will be pushed.
+#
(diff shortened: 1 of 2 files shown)


.githooks/pre-push
```diff
@@ -0,0 +1,42 @@
...
+if [[ -n `git diff` ]]
+then
+  echo "Please run cargo fmt"
+  git reset -q --hard HEAD
+  exit 1
+fi
+
+cargo clippy
+
+if [[ -n `git diff` ]]
```
padenot, line 36: This is probably incorrect, I don't think `clippy` makes changes ?

ChunMinChang: `cargo clippy -- -D warnings || (echo "Please run 'cargo clippy'"; false);` should work. I am going to merge it and have a quick fix.

PR #67, opened 2020-03-25
Revise `destroy_cubeb_device_info`

ChunMinChang:
The `device_id`, `group_id`, and `vendor_name` can be `null` pointer if `get_device_uid`, `get_device_model_uid`, `get_device_manufacturer` fails, while `friendly_name` will always be assigned to a pointer pointing to a valid `CString`. As a result, the memory retrieve of `device_id`, `group_id`, and `vendor_name` should be done only when they are not `null` and we should always assert `friendly_name` points to a valid address.

@padenot : I made a mistake to ask you change the `destroy_cubeb_device_info`. Here is a fix.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -1641,21 +1641,25 @@ fn is_aggregate_device(device_info: &ffi::cubeb_device_info) -> bool {
 }
 
 fn destroy_cubeb_device_info(device: &mut ffi::cubeb_device_info) {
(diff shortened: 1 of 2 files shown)


padenot (approved):
Ok. I don't think the API guarantees that any of those point to valid memory, but if we want to make it easier for authors, we could mandate it, and it would made our code simpler. This will have to be done across backends however.

PR #68, opened 2020-03-26
Revise git hooks

ChunMinChang:
The git hook is introduced in #66 but it doesn't work for `cargo clippy` check. Here is the fix.


padenot (approved):
Much better indeed, thanks!

PR #71, opened 2020-03-30
Fix the resampler-calling code to use the correct semantic

padenot:
The resampler copies all the frames into its internal buffer regardless
of the value it returns. Because we need to keep track of the buffer
depths in this backend to avoid growing the latency between input and
output, we only give the resampler enough data to work with.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -676,15 +676,17 @@ extern "C" fn audiounit_output_callback(
             // Else if the input has buffered a lot already because the output started late, we
             // need to trim the input buffer
             if prev_frames_written == 0 && buffered_input_frames > input_frames_needed as usize {
-                let samples_to_pop =
-                    (buffered_input_frames - input_frames_needed as usize) * input_channels;
+                let frames_to_pop = buffered_input_frames - input_frames_needed as usize;
+                let samples_to_pop = frames_to_pop * input_channels;
                 stm.core_stream_data
                     .input_linear_buffer
                     .as_mut()
                     .unwrap()
                     .pop(samples_to_pop);
                 stm.frames_read
-                    .fetch_sub((samples_to_pop / input_channels) as usize, Ordering::SeqCst);
+                    .fetch_sub(frames_to_pop as usize, Ordering::SeqCst);
+
+                cubeb_log!("Dropping {} frames in input buffer.", frames_to_pop);
             }
 
             if input_frames_needed > buffered_input_frames {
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -685,6 +683,8 @@ extern "C" fn audiounit_output_callback(
                     .pop(samples_to_pop);
                 stm.frames_read
                     .fetch_sub((samples_to_pop / input_channels) as usize, Ordering::SeqCst);
+
+                cubeb_log!("Dropping {} frames in input buffer.", samples_to_pop / input_channels);
```
ChunMinChang: nit: Declaring an avariable `let dropped_frames = samples_to_pop / input_channels` can avoid doing this caculation twice I guess.

src/backend/mod.rs
```diff
@@ -710,28 +709,18 @@ extern "C" fn audiounit_output_callback(
...
-                / stm.core_stream_data.input_desc.mChannelsPerFrame as usize;
-            cubeb_logv!("Total input frames: {}", input_frames);
             (
                 stm.core_stream_data
                     .input_linear_buffer
                     .as_mut()
                     .unwrap()
                     .as_mut_ptr(),
-                input_frames as i64,
+                input_frames_needed as i64,
```
ChunMinChang, line 721: Is `input_frames` different than `input_frames_needed` here? It looks like they are same values at this point?

src/backend/mod.rs
```diff
@@ -710,28 +709,18 @@ extern "C" fn audiounit_output_callback(
...
             )
         } else {
             (ptr::null_mut::<c_void>(), 0)
         };
 
-        // Call user callback through resampler.
-        assert!(!output_buffer.is_null());
```
ChunMinChang, line 734: Would this be checked in *cubeb_resampler*?

src/backend/mod.rs
```diff
@@ -453,11 +453,9 @@ extern "C" fn audiounit_input_callback(
         };
 
         // Advance input frame counter.
-        stm.frames_read
-            .fetch_add(input_frames as usize, atomic::Ordering::SeqCst);
+        stm.frames_read.fetch_add(input_frames as usize, atomic::Ordering::SeqCst);
```
ChunMinChang: I am not sure if this line needs to be re-formatted. The code style stays the same on my end when `cargo fmt --version`/`rustfmt --version` is either *rustfmt 1.4.11-stable (9eb4b564 2020-01-29)* or *rustfmt 1.4.12-nightly (9f53665f 2020-02-10)*.

Are you using`rustfmt.toml`?

src/backend/mod.rs
```diff
@@ -453,11 +453,9 @@ extern "C" fn audiounit_input_callback(
...
 
-        cubeb_logv!(
-            "({:p}) input: buffers {}, size {}, channels {}, rendered frames {}, total frames {}.",
+        cubeb_logv!("({:p}) input: buffers {}, size {}, channels {}, rendered frames {}, total frames {}.",
```
ChunMinChang: The style stays the same on my end.

src/backend/mod.rs
padenot: `input_frames` is the number of frames there are in the `input_linear_buffer`. `input_frames_needed` is the number of frames we need here for one callback.

src/backend/mod.rs
padenot: This is an invariant of the API: if we are in an output callback, then there is an output buffer.

padenot:
I've pushed an update with comments addressed and ran `cargo fmt`.

ChunMinChang:
> I've pushed an update with comments addressed and ran `cargo fmt`.

Is it supposed to have some differences between https://github.com/ChunMinChang/cubeb-coreaudio-rs/commit/55fd4bf0ca31febba9f59fb20d7c3d5bd8237fcf and https://github.com/ChunMinChang/cubeb-coreaudio-rs/commit/d49a6c6fe8a6a8fe5632f1a70a8464921d37114b? The only difference I see is the commit-id.

The style-check still fails on Travis CI server: https://travis-ci.org/github/ChunMinChang/cubeb-coreaudio-rs/jobs/668918572

padenot:
Yes, pushed the wrong one it seems like.

PR #74, opened 2020-04-01
Replace the custom input linear buffer by something based on ringbuf.

padenot:
This very similar to what we're doing in the pulse backend, we can
deduplicate after the repo merge.


ChunMinChang (approved):
LGTM

PR #76, opened 2020-04-02
Revise prepush when deleting branch

ChunMinChang:
When deleting a branch, there is no need to run `cargo fmt` and `cargo clippy`, so we should skip those checks. On the other hand, we should prevent the main branch from being deleted, so a check is added to prepush. In addition, the indent of all the scripts are updated to 4 spaces.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -26,12 +26,19 @@ IS_DESTRUCTIVE="\-\-delete|\-f"
 PROTECTED_BRANCH="trailblazer"
 CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 
+WILL_DELETE_PROTECTED_BRANCH=":$PROTECTED_BRANCH"
+
 if [[ $PUSH_COMMAND =~ $IS_DESTRUCTIVE ]]; then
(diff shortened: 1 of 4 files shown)


.githooks/pre-push
```diff
@@ -16,8 +16,30 @@
...
+PUSH_COMMAND=$(ps -ocommand= -p $PPID)
+
+IS_DESTRUCTIVE="\-\-delete|\-f"
+
+PROTECTED_BRANCH="trailblazer"
+CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+
+WILL_DELETE_PROTECTED_BRANCH=":$PROTECTED_BRANCH"
+
+if [[ $PUSH_COMMAND =~ $IS_DESTRUCTIVE ]]; then
```
padenot, line 31: `=~` is not valid `sh`. It's ok to use it, but change the shebang to `/bin/bash`. Surely any machine that is a source control server has `bash` and not only `sh`.

padenot (approved):

.githooks/pre-push
ChunMinChang: Good catch! thanks!

PR #79, opened 2020-04-07
Always pass all of the buffered input in the resampler

padenot:
As discussed on chat.mozilla.org.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -606,7 +606,10 @@ extern "C" fn audiounit_output_callback(
                 cubeb_log!("Dropping {} frames in input buffer.", popped_samples);
             }
 
-            if input_frames_needed > buffered_input_frames {
+            if input_frames_needed > buffered_input_frames
+                && (stm.switching_device.load(Ordering::SeqCst)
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -617,19 +619,18 @@ extern "C" fn audiounit_output_callback(
                     stm.core_stream_data.stm_ptr,
                     if stm.frames_read.load(Ordering::SeqCst) == 0 {
                         "input hasn't started,"
-                    } else if stm.switching_device.load(Ordering::SeqCst) {
-                        "device switching,"
                     } else {
-                        "drop out,"
```
ChunMinChang, line 623: Should we handle the *drop out* case? Not quite sure what the "drop out" originally means though. Is it the case cause the click?

ChunMinChang (commented):
Could you apply the `cargo fmt` so this can be tested on the Travis CI server?

src/backend/mod.rs
padenot: I means we're missing input frames, but that's handled by the resampler now.

PR #80, opened 2020-04-08
Update core-foundation-sys

kvark:
This is a strawman PR: I don't know anything about this library, in particular - about whether the dependency is public or not. I'm trying to align Gecko to the same `core-foundation-sys` and this would let us drop 0.6 from third-party dependencies (as a part of https://phabricator.services.mozilla.com/D70140)

### DIFF
--- a/coreaudio-sys-utils/Cargo.toml
+++ b/coreaudio-sys-utils/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "coreaudio-sys-utils"
-version = "0.1.1"
+version = "0.1.0"
 authors = ["Chun-Min Chang <chun.m.chang@gmail.com>"]
 edition = "2018"
 license = "ISC"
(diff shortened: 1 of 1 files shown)


coreaudio-sys-utils/Cargo.toml
```diff
@@ -1,12 +1,12 @@
 [package]
 name = "coreaudio-sys-utils"
-version = "0.1.0"
+version = "0.1.1"
```
ChunMinChang: I am not sure why `coreaudio-sys-utils` needs to be updated to `0.1.1`

ChunMinChang (commented):
Thanks for the fix! LGTM for the `core-foundation-sys` update.

However, I don't think I follow why we need to update `coreaudio-sys-utils`. `coreaudio-sys-utils` is only used internally for this crate.

coreaudio-sys-utils/Cargo.toml
kvark: Indeed, this should not be here. The dependency is specified as github path, we don't have to publish that.

kvark:
Thanks for taking a look. Is it better now?

ChunMinChang (approved):
Thanks! LGTM!

kvark:
@ChunMinChang is it safe to point Gecko to latest git revision here? Any testing, precautions that need to be done beforehand?

PR #82, opened 2020-04-13
Update to cubeb-backend 0.6.3.

kinetiknz:
For [BMO 1628132](https://bugzilla.mozilla.org/show_bug.cgi?id=1628132).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ crate-type = ["staticlib", "rlib"]
 atomic = "0.4"
 bitflags = "1.0"
 coreaudio-sys-utils = { path = "coreaudio-sys-utils" }
-cubeb-backend = "0.6"
+cubeb-backend = "0.6.3"
 float-cmp = "0.6"
 libc = "0.2"
 lazy_static = "1.2"


ChunMinChang (approved):
The update looks good to me. 

BTW, I wonder if we could pass the gecko build without changing the version. The `cubeb-*` crates could be updated to `0.6.3` by `cargo update -p cubeb-backend && ./mach vendor rust --ignore-modified`. It still meets the minimal version requirements `0.6` here.

ChunMinChang:
It seems somehow it makes the `cargo clippy` aware of the styling issues in this crates now. Would you like to fix them as well in this PR? I am ok to get them fixed in another PR.

kinetiknz:
> BTW, I wonder if we could pass the gecko build without changing the version. The `cubeb-*` crates could be updated to `0.6.3` by `cargo update -p cubeb-backend && ./mach vendor rust --ignore-modified`. It still meets the minimal version requirements `0.6` here.

That might be easier. Not sure what the best practice is in this situation.

> It seems somehow it makes the `cargo clippy` aware of the styling issues in this crates now. Would you like to fix them as well in this PR? I am ok to get them fixed in another PR.

I'd rather leave them alone, they're potentially not simple to fix (I haven't looked at the code to check) unless you're fine with just ignoring `clippy::cognitive-complexity` on those functions - it seems like more of a taste issue than a real lint in my opinion.

ChunMinChang:
> I'd rather leave them alone, they're potentially not simple to fix (I haven't looked at the code to check) unless you're fine with just ignoring `clippy::cognitive-complexity` on those functions - it seems like more of a taste issue than a real lint in my opinion.

It is. I am ok to use `#[allow(clippy::cognitive_complexity)]` but we can do this in another PR and get this merged first!

PR #85, opened 2020-04-14
Lower `cognitive_complexity` of `get_device_group_id`

ChunMinChang:
A simple style revise to make `cargo clippy` happy.


padenot:
Ok, but I don't think it's particularly useful to spend time on things like this.

padenot (approved):

ChunMinChang:
This is just an precaution for keeping [the Broken Windows](https://en.wikipedia.org/wiki/Broken_windows_theory) out of the sight.

PR #86, opened 2020-04-15
Don't really push silence to the ring buffer

padenot:
This implements your idea, only by removing code it turns out.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -399,12 +399,9 @@ extern "C" fn audiounit_input_callback(
             // For now state that no error occurred and feed silence, stream will be
             // resumed once reinit has completed.
             cubeb_logv!(
-                "({:p}) input: reinit pending feeding silence instead",
+                "({:p}) input: reinit pending, output will pull silence instead",
                 stm.core_stream_data.stm_ptr
             );
-            let elements =
-                (input_frames * stm.core_stream_data.input_desc.mChannelsPerFrame) as usize;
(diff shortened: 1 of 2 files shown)


padenot:
I've tested it on a few devices on my macbook 2018: built-in speakers/mic, an hd25 plugged in the TRRS socket(it doesn't have a mic), and a Sony WH-1000 xm3, that has a mic, via bluetooth, and trying all sorts of combinations and switches.

src/backend/mod.rs
```diff
@@ -611,8 +608,6 @@ extern "C" fn audiounit_output_callback(
                     || stm.frames_read.load(Ordering::SeqCst) == 0)
             {
                 let silent_frames_to_push = input_frames_needed - buffered_input_frames;
-                let silent_samples_to_push = silent_frames_to_push * input_channels;
-                input_buffer_manager.push_silent_data(silent_samples_to_push);
```
ChunMinChang, line 615: Can you add some comments here, or above `get_linear_data`, or `cubeb_log!` below to not that the silent data is appended silently in `get_linear_data`?

ChunMinChang (approved):
LGTM! 

r+ with some comments added.

src/backend/mod.rs
padenot: Yep did both.

ChunMinChang:
Add a reference: This fixes [BMO 1629839](https://bugzilla.mozilla.org/show_bug.cgi?id=1629839).

PR #88, opened 2020-04-24
Add an implementation for `cubeb_stream_get_input_latency`

padenot:
MacOS implementation for https://github.com/kinetiknz/cubeb/pull/583 (this can't build right now we need to merge and bump `cubeb-rs`).

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -344,6 +344,10 @@ extern "C" fn audiounit_input_callback(
     assert!(!user_ptr.is_null());
     let stm = unsafe { &mut *(user_ptr as *mut AudioUnitStream) };
 
+    let input_latency_frames = compute_input_latency(&stm, unsafe { (*tstamp).mHostTime });
+    stm.total_input_latency_frames
+        .store(input_latency_frames, Ordering::SeqCst);
+
     if stm.shutdown.load(Ordering::SeqCst) {
         cubeb_log!("({:p}) input shutdown", stm as *const AudioUnitStream);
         return NO_ERR;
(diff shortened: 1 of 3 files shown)


src/backend/mod.rs
```diff
@@ -2822,8 +2843,8 @@ impl<'ctx> CoreStreamData<'ctx> {
                 return Err(Error::error());
             }
 
-            stream.current_latency_frames.store(
-                get_presentation_latency(self.output_device.id, DeviceType::OUTPUT),
+            stream.current_output_latency_frames.store(
+                get_fixed_latency(self.output_device.id, DeviceType::OUTPUT),
```
ChunMinChang, line 2852: Do we need to apply the same manner for`current_input_latency_frames`?

src/backend/mod.rs
```diff
@@ -501,7 +505,24 @@ fn compute_output_latency(stm: &AudioUnitStream, host_time: u64) -> u32 {
...
+    } else {
+        now - audio_input_time
+    };
+
+    const NS2S: u64 = 1_000_000_000;
+    // The total input latency is the timestamp difference + the stream latency +
+    // the hardware latency.
+    let input_hw_rate = stm.core_stream_data.input_hw_rate as u64;
+    (input_latency_ns * input_hw_rate / NS2S
+        + stm.current_input_latency_frames.load(Ordering::SeqCst) as u64) as u32
```
ChunMinChang, line 525: Is `stm.current_input_latency_frames` always `0`? Should the device-latency and stream-latency be added to `stm.current_input_latency_frames`?

src/backend/mod.rs
```diff
@@ -3445,6 +3471,21 @@ impl<'ctx> StreamOps for AudioUnitStream<'ctx> {
     fn latency(&mut self) -> Result<u32> {
         Ok(self.total_output_latency_frames.load(Ordering::SeqCst))
     }
+    #[cfg(target_os = "ios")]
+    fn input_latency(&mut self) -> Result<u32> {
+        Err(not_supported())
+    }
+    #[cfg(not(target_os = "ios"))]
+    fn input_latency(&mut self) -> Result<u32> {
+        let user_rate = self.core_stream_data.output_stream_params.rate();
```
ChunMinChang: Does `input_latency` work for input-only stream? IIRC, `output_stream_params.rate()` will be `0` if it's a input-only stream. In that case, this will return `0`. Is it expected?

ChunMinChang:
LGTM! We should try to run the test after cubeb-rs is bumped!

src/backend/mod.rs
padenot: Yes my mistake.

src/backend/mod.rs
padenot: Yeah same, I couldn't compile and test so I didn't see my mistake.

padenot:
I've now taken some time to test this by forcing paths in `Cargo.toml`. I have similar number with https://padenot.github.io/roundtrip-latency-tester/ and with this API (by summing input and output latency of course), so it works well.

PR #97, opened 2020-06-04
Fix frames_read value

Koalab99:
I couldn't test it, but it should fix some problem

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -624,11 +624,10 @@ extern "C" fn audiounit_output_callback(
             // need to trim the input buffer
             if prev_frames_written == 0 && buffered_input_frames > input_frames_needed as usize {
                 input_buffer_manager.trim(input_frames_needed * input_channels);
-                let popped_samples =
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -624,11 +624,10 @@ extern "C" fn audiounit_output_callback(
             // need to trim the input buffer
             if prev_frames_written == 0 && buffered_input_frames > input_frames_needed as usize {
                 input_buffer_manager.trim(input_frames_needed * input_channels);
-                let popped_samples =
-                    ((buffered_input_frames - input_frames_needed) * input_channels) as usize;
-                stm.frames_read.fetch_sub(popped_samples, Ordering::SeqCst);
+                let popped_frames = buffered_input_frames - input_frames_needed as usize;
+                stm.frames_read.fetch_sub(popped_frames, Ordering::SeqCst);
```
ChunMinChang, line 628: Good catch, thanks!

ChunMinChang (approved):

PR #98, opened 2020-06-09
Return a position even when not latency info are available

padenot:
The behaviour changed in
https://github.com/ChunMinChang/cubeb-coreaudio-rs/commit/38e0c0ffebf2902e310f7ed35e90cfe92075cf5c


achronop (approved):
Looks good, thanks

PR #101, opened 2020-07-21
Fix clock reporting by using the correct latency number

padenot:
It's wrong, but it matters more in some scenarii than others:

Testing on my mbp 2018 with the built-in output device the device + stream latency is 17 frames, and the total latency (taking into account the timestamp provided in the callback) is 544. This was at 48kHz and this translates to 300us and 11.3ms respectively. Quite a big difference, but not really noticeable for A/V sync when not looking for it, I assume.

For a bluetooth device (Sony WH1000-XM3) on the same machine, it's 7872 frames of latency for the device, and 9148 total, still at 48kHz, this translates to respectively 164ms and 190ms, which is quite a lot (a bit less than two frames at 60fps, a bit less than a frame at 30fps). 

Not the end of the world, but still better. This [mailing list post](https://lists.apple.com/archives/coreaudio-api/2017/Jul/msg00035.html) has background on this.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -3468,14 +3468,14 @@ impl<'ctx> StreamOps for AudioUnitStream<'ctx> {
         Err(Error::not_supported())
     }
     fn position(&mut self) -> Result<u64> {
-        let current_output_latency_frames =
-            u64::from(self.current_output_latency_frames.load(Ordering::SeqCst));
+        let total_output_latency_frames =
+            u64::from(self.total_output_latency_frames.load(Ordering::SeqCst));
         let frames_played = self.frames_played.load(Ordering::SeqCst);
-        if current_output_latency_frames != 0 {
-            let position = if current_output_latency_frames > frames_played {
+        if total_output_latency_frames != 0 {
+            let position = if total_output_latency_frames > frames_played {
                 0
             } else {
-                frames_played - current_output_latency_frames
+                frames_played - total_output_latency_frames
             };
             return Ok(position);
         }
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -500,7 +500,7 @@ fn compute_output_latency(stm: &AudioUnitStream, host_time: u64) -> u32 {
     let audio_output_time = host_time_to_ns(host_time);
     let output_hw_rate = stm.core_stream_data.output_hw_rate as u64;
     let fixed_latency_ns =
-        (stm.current_output_latency_frames.load(Ordering::SeqCst) as u64 * NS2S) / output_hw_rate;
+        (stm.output_device_latency_frames.load(Ordering::SeqCst) as u64 * NS2S) / output_hw_rate;
     let total_output_latency_ns = if audio_output_time < now {
         0
```
ChunMinChang: I wonder if `total_output_latency_ns` should be changed to:

```rust
let total_output_latency_ns = fixed_latency_ns + audio_output_time.checked_sub(now).unwrap_or(0)
```

so the `total_output_latency_ns` has a larger-than-0 value once the output callback is fired.

Otherwise, the `total_output_latency_ns` will only have a larger-than-0 value once `audio_output_time >= now`.

ChunMinChang (approved):

src/backend/mod.rs
padenot: Yes, good idea.

ChunMinChang:
Add some references
- https://lists.apple.com/archives/coreaudio-api/2005/Dec/msg00055.html
- https://lists.apple.com/archives/coreaudio-api/2020/May/msg00004.html

PR #107, opened 2020-07-27
Fix more things about the clock reporting, and add a test

### LINKED ISSUE #106: `position` keeps growing even when stream is stopped
[The failing test](https://treeherder.mozilla.org/logviewer.html#/jobs?job_id=310990910&repo=autoland&lineNumber=1803) in [BMO 1654699](https://bugzilla.mozilla.org/show_bug.cgi?id=1654699#c9) indicates the `position` keeps growing when stream is already stopped.

I guess the reason is that `interpolated_frames` is increased unlimitedly.

One approach is to add `outframes` to `OutputCallbackTimingData` and use it to create a bound for `interpolated_frames` (e.g., `interpolated_frames = min(interpolated_frames, outframes)`). Or we can apply the same manner on WASAPI backend for this case.

padenot:


src/backend/tests/tone.rs
```diff
@@ -36,10 +44,125 @@ fn test_dial_tone() {
...
+                        }
+                    }
+                    State::Paused => {
+                        // The cubeb_stream_stop call above should synchrously stop the callbacks,
+                        // hence the clock, the assert below must always holds, modulo the client
+                        // side interpolation.
+                        assert!(
+                            position == prev_position
+                                || position - prev_position
+                                    < closure.buffer_size.load(Ordering::SeqCst) as u64
```
ChunMinChang: As we discussed offline, this should be `position - prev_position <= closure.buffer_size.load(Ordering::SeqCst) as u64`

ChunMinChang (approved):
Looks good to me! Thanks for fixing the issue!

PR #110, opened 2020-10-06
Destroy newly created aggregate device if its initialization fails

ChunMinChang:
If the created aggregate device (via `create_blank_device_sync`) fails during its initialization (e.g., `set_sub_devices_sync`, `set_master_device`, ...), the newly created aggregate device should be destroyed.

https://github.com/ChunMinChang/cubeb-coreaudio-rs/blob/91a7bb50198a9b7578d98d56e9fd32f257abb41f/src/backend/aggregate_device.rs#L37-L42

### DIFF
--- a/src/backend/utils.rs
+++ b/src/backend/utils.rs
@@ -35,32 +35,25 @@ pub fn cubeb_sample_size(format: fmt) -> usize {
     }
 }
 
-pub struct Finalizer<F: FnOnce()> {
-    func: Option<F>,
-    commit: bool,
-}
+pub struct Finalizer<F: FnOnce()>(Option<F>);
 
 impl<F: FnOnce()> Finalizer<F> {
     pub fn dismiss(&mut self) {
-        self.commit = false;
+        let _ = self.0.take();
+        assert!(self.0.is_none());
     }
 }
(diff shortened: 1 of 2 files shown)


ChunMinChang:
The test fails until #111 is merged.

src/backend/utils.rs
```diff
@@ -35,18 +35,32 @@ pub fn cubeb_sample_size(format: fmt) -> usize {
...
         }
     }
 }
 
-pub fn finally<F: FnOnce()>(f: F) -> impl Drop {
-    Finalizer(Some(f))
+pub fn finally<F: FnOnce()>(f: F) -> Finalizer<F> {
+    Finalizer {
+        func: Some(f),
+        commit: true,
```
padenot: I'd have merged `commit` and `func`. If `func` is `Some` then run it. Otherwise, do nothing.

padenot (approved):
r+ with a suggestion if you want.

src/backend/utils.rs
ChunMinChang: Yeah, that's neat! I'll update the patch! Thanks!

PR #112, opened 2020-10-07
Call `audio_object_{add, remove}_property_listener` outside of `assert_eq`

ChunMinChang:
In `create_blank_device_sync`, `audio_object_{add, remove}_property_listener` is called within a `assert_eq!`, which may not be executed in optimized build.


ChunMinChang:
The test fails until #111 is merged.

padenot (approved):
I wonder if we could have a static analysis thing for this.

PR #116, opened 2020-10-21
Respect mach timebase when computing latencies.

mstange:
Fix for [bug 1665243](https://bugzilla.mozilla.org/show_bug.cgi?id=1665243)

This fixes a bug where incorrect position information would be returned
when compiling for Apple Silicon, where mach time values do not map 1:1
to nanoseconds.
See https://developer.apple.com/documentation/apple_silicon/addressing_architectural_differences_in_your_macos_code#3616875

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -494,28 +494,28 @@ fn host_time_to_ns(host_time: u64) -> u64 {
     rv as u64
 }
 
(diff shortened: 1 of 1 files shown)


SingingTree:
I think there's a typo in that last link and it should be https://developer.apple.com/documentation/apple_silicon/addressing_architectural_differences_in_your_macos_code#3616875

Edit: Also, thanks for figuring the root cause out and patching!

mstange:
Thanks, fixed.

PR #118, opened 2020-11-18
Bail out in `activate_clock_drift_compensation` if no enough sub devices

ChunMinChang:
This works around the crash in [BMO 1677766](https://bugzilla.mozilla.org/show_bug.cgi?id=1677766).

It's not clear why the number of the owning sub-devices in the created
aggregate device is less than 2, after setting sub devices from input
and output sides via `AggregateDevice::set_sub_devices_sync`. The
sub-devices should be set synchronously when the above function is
called. Maybe `kAudioObjectPropertyOwnedObjects` is not synchronous or
symmetric to `kAudioAggregateDevicePropertyFullSubDeviceList`, which is
the property set in `AggregateDevice::set_sub_devices_sync`.

If the sub devices are less than 2 (we should have at least one input
and one output) in `AggregateDevice::activate_clock_drift_compensation`,
the `AggregateDevice::new` should bail out with an error returned. Then
the cubeb stream can follow the fallback mechanism to create a duplex
stream without using an aggregate-device.

### DIFF
--- a/src/backend/aggregate_device.rs
+++ b/src/backend/aggregate_device.rs
@@ -15,7 +15,7 @@ pub struct AggregateDevice {
 pub enum Error {
     OS(OSStatus),
     Timeout(std::time::Duration),
-    LessDevicesThan2(usize),
+    LessThan2Devices(usize),
 }
 
 impl From<OSStatus> for Error {
@@ -32,7 +32,7 @@ impl From<std::time::Duration> for Error {
 
 impl From<usize> for Error {
     fn from(number: usize) -> Self {
-        Error::LessDevicesThan2(number)
+        Error::LessThan2Devices(number)
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/backend/aggregate_device.rs
```diff
@@ -15,6 +15,7 @@ pub struct AggregateDevice {
 pub enum Error {
     OS(OSStatus),
     Timeout(std::time::Duration),
+    LessDevicesThan2(usize),
```
padenot: it: lessthan2devices sounds a bit more natural.

padenot (approved):

src/backend/aggregate_device.rs
ChunMinChang: Yeah, that's better. I will update the patch later. Maybe I’ve been in lockdown too long and forget how to speak English.

PR #122, opened 2021-02-20
BMO 1692910: Avoid the race between `set_volume` and stream-reinitialization

ChunMinChang:
This solves the assertion-failed issue in [BMO 1692910](https://bugzilla.mozilla.org/show_bug.cgi?id=1692910).

This is a data-racing issue.  The assertion will be hit when the
1. `set_volume` is called on thread *A* at the same time when the stream is being reinitialized at thread *B*
2. `set_volume` is called on thread *A* just after the `self.core_stream_data.close()` is called on thread *B*, which will uninitialize and dispose the AudioUnit called in `set_volume` on thread *A*

This issue can be reproduced by operating `test_stream_tester()`  with the code added
in 8c1e3f91bd05643113d00ba93c965711f5ac3c10 and 32f840c338900515452f4f7cc2be6e97ddd21279, by following the commit message in 8c1e3f91bd05643113d00ba93c965711f5ac3c10. The code added in 8c1e3f91bd05643113d00ba93c965711f5ac3c10 in stream-reinitialization task can buy us some time to call `set_volume` when the stream is being reinitialized. That code is removed in bf40fb55d5ab4f7dcd1dc7fcd267520bd88ddfbf since it's just used to reproduce the issue and check the code in 601d37fbba3951929411cca8f335974097f6bd09 can fix the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,8 @@ It's used to verify our callbacks for minitoring the system devices work.
     - `d` to destroy a stream
     - `s` to start the created stream
     - `t` to stop the created stream
+    - `r` to register a device changed callback to the created stream
+    - `v` to set volume to the created stream
     - `q` to quit the test
   - It's useful to simulate the stream bahavior to reproduce the bug we found,
     with some modified code.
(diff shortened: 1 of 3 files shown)


kinetiknz (approved):
LGTM

PR #127, opened 2021-03-09
Rework `test_unplug_a_device_on_an_active_stream`

ChunMinChang:
When we create/add/plug a new device to Mac OS, the newly added device
is possible to be the new default device (depends on system setting?).
In `test_unplug_a_device_on_an_active_stream`, we didn't check if the
default device is the expected one. The created device in that function
could be the default device in some tests, and it could be non-default
device in some tests. We need to check if the created device is default
or non-default before running the tests. Otherwise we may use the wrong
device to run the tests.

### DIFF
--- a/src/backend/tests/device_change.rs
+++ b/src/backend/tests/device_change.rs
@@ -523,11 +523,46 @@ fn test_unplug_a_device_on_an_active_stream(
         return;
     }
 
+    let default_device_before_plugging = test_get_default_device(device_scope.clone()).unwrap();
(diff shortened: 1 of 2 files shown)


ChunMinChang:
This is caught by running github-action: https://github.com/ChunMinChang/cubeb-coreaudio-rs/runs/2069461472?check_suite_focus=true#step:5:586

kinetiknz (approved):
LGTM, thanks.

PR #128, opened 2021-03-09
Add GitHub Action workflows

### LINKED ISSUE #120: Travis CI free usage ends Dec 3; mozilla repos should switch to other CI platforms
Same issue as https://github.com/mozilla/cubeb-pulse-rs/issues/64

We should probably move from Travis CI to [CircleCI](https://circleci.com/)

ChunMinChang:
Circle-CI won't run its build and tests for the PR request, but we can use Github-Action to do this. Another benefit I found is that the Mac OS server for Github-Action has a built-in audio device (Soundflower, actually). All the device-related tests can be fully tested on the server.


nils-ohlmeier:
@ChunMinChang we ran across the same issue of no Circle-CI runs for PR's, but according to this tweet that is only a configuration issue: https://twitter.com/zmarkan/status/1367827983457779716

ChunMinChang:
> @ChunMinChang we ran across the same issue of no Circle-CI runs for PR's, but according to this tweet that is only a configuration issue: https://twitter.com/zmarkan/status/1367827983457779716

I've turned it on yesterday, but it doesn't seem to work. At least, I don't see it runs for this PR

Screen Shot 2021-03-08 at 9 29 42 PM

ChunMinChang:
I might need to do this: https://circleci.com/docs/2.0/enable-checks/ as well to make it work for PR. I'll try tomorrow. However, it's worth using Github-Action since the Github-Action server has a built-in audio device, while the Circle-CI server has not.

kinetiknz:
I used GitHub Actions for the other cubeb repos, also finding Soundflower available for macOS tests.

The actions here look good, but as I understand it actions-rs and other external actions are generally not available inside the mozilla GitHub org for security reasons.  The Settings -> Actions page for each repo shows:
image

[cubeb-pulse-rs](https://github.com/mozilla/cubeb-pulse-rs/blob/dev/.github/workflows/build.yml) has a fairly simple working build action without actions-rs, if you'd like to base an action on that.

ChunMinChang:
@kinetiknz : I've updated the GitHub Actions setting. Would you mind taking a look?

kinetiknz (approved):
LGTM

.github/workflows/test.yml
```diff
@@ -0,0 +1,45 @@
...
+        submodules: recursive
+
+    - name: Install Rust
+      run: rustup toolchain install ${{ matrix.rust }} --profile minimal --component rustfmt clippy
+        
+    - name: Setup
+      env:
+          MATRIX_RUST: ${{ matrix.rust }}
+      run: |
+        echo matrix.rust=$MATRIX_RUST
```
kinetiknz: The action logs will show the actual value of `${{ matrix.rust }}` and other substitutions where used, so logging the vars is probably not needed.

ChunMinChang: Good point. I've updated the file.

PR #132, opened 2021-07-22
BMO 1721496 - Append silence frames to when switching from A2DP to HFP/HSP

ChunMinChang:
This fixes [BMO 1721496](https://bugzilla.mozilla.org/show_bug.cgi?id=1721496)

No silence frame is fed to the resampler when the duplex stream runs on a
Bluetooth device when the device is switching from A2DP to HFP/HSP profile,
which leads to hitting an assertion [1] within cubeb resampler.

When the duplex cubeb stream runs on a Bluetooth device and the device
is switching from A2DP to HFP/HSP profile, the `input_buffer_manager`
has no input data to be appended [2] in `audiounit_input_callback`. And
then when `audiounit_output_callback` runs, the `buffered_input_frames`
is `0` but `input_frames_needed` is a positive integer greater than `0`.
However, the `input_frames` in this case is equal to
`buffered_input_frames` in current implementation since neither 
`switching_device` nor `frames_read > 0` is `true`. This is wrong. The
`input_frames` should be `input_frames_needed` rather than `0`.

When the above case happens in the first `audiounit_output_callback`
call, it makes the cubeb resampler hit the assertion in [1]. The `data_`
[3] in the `internal_input_buffer` [4] is `nullptr` when the resampler
is created (`passthrough_resampler` in my case). If the `length` in the
first `push*` [5] is `0`, then the `reserve` [6] won't be called so `data_`
remains `nullptr`, which makes the resampler hit the assertion in [1]

We should append the silence frames to the resampler in this case not
only to avoid hitting the assertion but also to append the silence
frames correctly when we need.

[1] https://github.com/mozilla/cubeb/blob/b608f590249d5821c91b3ea83e394f0fd839f5b9/src/cubeb_utils.h#L31
[2] https://github.com/mozilla/cubeb-coreaudio-rs/blob/8c9bb1e74507b1fcac849c91fc6eb6f86d44692d/src/backend/mod.rs#L409
[3] https://github.com/mozilla/cubeb/blob/b608f590249d5821c91b3ea83e394f0fd839f5b9/src/cubeb_utils.h#L115
[4] https://github.com/mozilla/cubeb/blob/b608f590249d5821c91b3ea83e394f0fd839f5b9/src/cubeb_resampler_internal.h#L112
[5] https://github.com/mozilla/cubeb/blob/b608f590249d5821c91b3ea83e394f0fd839f5b9/src/cubeb_utils.h#L196
[6] https://github.com/mozilla/cubeb/blob/b608f590249d5821c91b3ea83e394f0fd839f5b9/src/cubeb_utils.h#L172

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -416,12 +416,13 @@ extern "C" fn audiounit_input_callback(
             let elements =
                 (input_frames * stm.core_stream_data.input_desc.mChannelsPerFrame) as usize;
             input_buffer_manager.push_data(input_buffer_list.mBuffers[0].mData, elements);
+
+            // Advance input frame counter.
+            stm.frames_read
+                .fetch_add(input_frames as usize, atomic::Ordering::SeqCst);
+
             ErrorHandle::Return(status)
         };
-
-        // Advance input frame counter.
-        stm.frames_read
-            .fetch_add(input_frames as usize, atomic::Ordering::SeqCst);
 
         cubeb_logv!(
             "({:p}) input: buffers {}, size {}, channels {}, rendered frames {}, total frames {}.",
@@ -637,7 +638,7 @@ extern "C" fn audiounit_output_callback(
                 silent_frames_to_push
             );
             stm.frames_read
-                .fetch_add(input_frames_needed, Ordering::SeqCst);
+                .fetch_add(silent_frames_to_push, Ordering::SeqCst);
             input_frames_needed
         } else {
             buffered_input_frames
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -619,23 +619,20 @@ extern "C" fn audiounit_output_callback(
             cubeb_log!("Dropping {} frames in input buffer.", popped_frames);
         }
 
-        let input_frames = if input_frames_needed > buffered_input_frames
-            && (stm.switching_device.load(Ordering::SeqCst)
-                || stm.reinit_pending.load(Ordering::SeqCst)
-                || stm.frames_read.load(Ordering::SeqCst) == 0)
-        {
+        let input_frames = if input_frames_needed > buffered_input_frames {
             // The silent frames will be inserted in `get_linear_data` below.
             let silent_frames_to_push = input_frames_needed - buffered_input_frames;
             cubeb_log!(
-                "({:p}) Missing Frames: {} will append {} frames of input silence.",
+                "({:p}) Missing Frames: {}, will append {} frames of input silence.",
                 stm.core_stream_data.stm_ptr,
                 if stm.frames_read.load(Ordering::SeqCst) == 0 {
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -620,6 +620,7 @@ extern "C" fn audiounit_output_callback(
 
         let input_frames = if input_frames_needed > buffered_input_frames
             && (stm.switching_device.load(Ordering::SeqCst)
+                || stm.reinit_pending.load(Ordering::SeqCst)
                 || stm.frames_read.load(Ordering::SeqCst) == 0)
```
ChunMinChang: @padenot: Do you remember why we need to add `(stm.switching_device.load(Ordering::SeqCst) || stm.reinit_pending.load(Ordering::SeqCst) || stm.frames_read.load(Ordering::SeqCst) == 0)` in this `if` check? It would be simpler to use `input_frames_needed > buffered_input_frames` only here.

ChunMinChang:
#131 should fix the test failure here

src/backend/mod.rs
padenot: That might work, yes. It used to be that the `buffer_manager` wasn't capable of under-running on its own, so we needed to be more careful.

ChunMinChang: Thanks! I'll remove these checks.

padenot: Please do a thorough manual test, this is hard to catch.

src/backend/mod.rs
```diff
@@ -618,25 +619,25 @@ extern "C" fn audiounit_output_callback(
...
                 } else {
-                    assert!(stm.switching_device.load(Ordering::SeqCst));
-                    "device switching,"
+                    "not enough buffered frames"
                 },
                 silent_frames_to_push
             );
             stm.frames_read
-                .fetch_add(input_frames_needed, Ordering::SeqCst);
+                .fetch_add(silent_frames_to_push, Ordering::SeqCst);
```
ChunMinChang: I guess this should be `silent_frames_to_push` since we've updated the `frames_read` in `audiounit_input_callback`. Otherwise, the `frames_read` will be *K+K+N*, where *K* is the `input_frames`, which is `buffered_input_frames` in most of the cases (when `reinit_pending` is false here),`N` is the `silent_frames_to_push` and `K +N` is `input_frames_needed`.

It seems to me `stm.frames_read` is a footgun. It doesn't serve for anything, just keep tracking how many input frames have been advanced. Could we remove `frames_read`, or at least merge it in `input_buffer_manager`? What do you think?

ChunMinChang: I've updated the code and make `frames_read` follows `input_buffer_manager.get_linear_data(FRAMES)` and always add `FRAMES` to `frames_read`

padenot (approved):

PR #135, opened 2021-08-25
BMO 1727598 - Fix audio cut

ChunMinChang:
This reverts commit 246b7873761390d30ed32b37419350f2c2e72572 in #133. This should fix [BMO 1727598](https://bugzilla.mozilla.org/show_bug.cgi?id=1727598). 

After applying this patch, the silence frames will only be appended when the resampler asking for more input frames than what we have *and* the device is switching or the stream is re-initializing. If the resampler asking for more input frames than what we have but the device is not switching or the stream is not re-initializing, now we count on [cubeb resmapler](https://github.com/mozilla/cubeb/blob/6ce95962c8bb1442a725cbacdc77d5d8cbce43ad/src/cubeb_resampler.h)'s help.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -615,20 +615,23 @@ extern "C" fn audiounit_output_callback(
             cubeb_log!("Dropping {} frames in input buffer.", popped_frames);
         }
 
-        let input_frames = if input_frames_needed > buffered_input_frames {
+        let input_frames = if input_frames_needed > buffered_input_frames
+            && (stm.switching_device.load(Ordering::SeqCst)
+                || stm.reinit_pending.load(Ordering::SeqCst)
+                || stm.frames_read.load(Ordering::SeqCst) == 0)
+        {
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -615,20 +615,23 @@ extern "C" fn audiounit_output_callback(
...
-                    "input hasn't started"
+                    "input hasn't started,"
                 } else if stm.switching_device.load(Ordering::SeqCst) {
-                    "device switching"
-                } else if stm.reinit_pending.load(Ordering::SeqCst) {
-                    "reinit pending"
+                    "device switching,"
                 } else {
-                    "not enough buffered frames"
+                    assert!(stm.reinit_pending.load(Ordering::SeqCst));
```
padenot: This doesn't have to be true at this point (https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use). The rest is good.
```suggestion

```

padenot (approved):

PR #136, opened 2021-09-30
Choose the size of the input ring buffer based on the stream parameters

padenot:
This fixes [BMO#1725203](https://bugzilla.mozilla.org/show_bug.cgi?id=1725203).

The method `discard` of `ringbuf` wasn't available in the version we had.


src/backend/mod.rs
```diff
@@ -2596,7 +2596,10 @@ impl<'ctx> CoreStreamData<'ctx> {
                 return Err(Error::error());
             }
 
-            self.input_buffer_manager = Some(BufferManager::new(self.input_stream_params.format()));
+            self.input_buffer_manager = Some(BufferManager::new(
+                &self.input_stream_params,
+                stream.latency_frames,
```
ChunMinChang: https://github.com/mozilla/cubeb-coreaudio-rs/blob/4ee3b283182926f9a8bb31d603219ff2db70d2fd/src/backend/mod.rs#L2171-L2182

https://github.com/mozilla/cubeb-coreaudio-rs/blob/4ee3b283182926f9a8bb31d603219ff2db70d2fd/src/backend/mod.rs#L2208-L2214

Note that the `stream.latency_frames` here is actually a *context-wise* value (same value in whole cubeb context) instead of the value passed in `stream_init` due to the issue mentioned [here](https://github.com/mozilla/cubeb-coreaudio-rs/blob/4ee3b283182926f9a8bb31d603219ff2db70d2fd/README.md?plain=1#L109). See [here](https://source.chromium.org/chromium/chromium/src/+/master:media/audio/mac/audio_manager_mac.cc;l=992-999;drc=f8ad3972f72cc90b4c7e20c9fdfb49b9edb31f72) for more details.

If we only support one output device and the *latency* is from `min_latency`, I would expect we always get the same value (do we support two output device in same cubeb context?):

https://github.com/mozilla/cubeb-coreaudio-rs/blob/4ee3b283182926f9a8bb31d603219ff2db70d2fd/src/backend/mod.rs#L2080-L2086

So I guess it's ok to use *global* latency here since the buffer-size is calculated by the *output channel count* as well.

ChunMinChang (approved):
LGTM, thanks!

My only concern is that the stream latency is a per-context value instead of a per-stream value. If we only support one output device per cubeb context, this should be fine.

src/backend/mod.rs
padenot: > If we only support one output device and the _latency_ is from `min_latency`, I would expect we always get the same value (do we support two output device in same cubeb context?):

We do. I think I'm just going to make this `SAFE_MAX_LATENCY * input_channel_count * sizeof(sample-type)`.

padenot:
Rebased and used a constant to implement what I said above.

ChunMinChang (approved):
LGTM, thanks!

PR #137, opened 2021-10-01
Fix nightly clippy warnings

ChunMinChang:
Fix a couple of warnings when building with rust nightly.


ChunMinChang:
This should solve the build error in #136: https://github.com/mozilla/cubeb-coreaudio-rs/pull/136/checks?check_run_id=3756279581

PR #141, opened 2021-11-09
BMO 1737833 - Return error when reinit fails if input is not defined

ChunMinChang:
If the stream is output-only, or its input device is system-default, its
reinit failure can lead to hitting an assertion in set_volume since the
program keeps running as the error doesn't occur. We should return an
error instead.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -3294,7 +3294,7 @@ impl<'ctx> AudioUnitStream<'ctx> {
             e
         })?;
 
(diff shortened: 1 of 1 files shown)


ChunMinChang:
It surprises me that we didn't take care of it for years: https://searchfox.org/mozilla-central/rev/7241c1f9e94377b70b49cf702264269779c5c322/media/libcubeb/src/cubeb_audiounit.cpp#853-866. I should have caught it when rewritting

padenot (approved):

PR #143, opened 2021-11-09
Fix clippy warnings

ChunMinChang:
Fix a couple of warnings found in nightly


ChunMinChang:
It looks like `clippy::non_send_fields_in_send_ty` is nightly-only now.

kinetiknz (approved):
LGTM.  I ran into the same issue in https://github.com/mozilla/audioipc-2/pull/131; dealt with the nightly-only issue with `#[allow(unknown_lints)] // non_send_fields_in_send_ty is Nightly-only as of 2021-11-29.` for now.

PR #152, opened 2022-03-01
Do not reinitialize audio stream if its input device isn't changed

### LINKED ISSUE #149: Do not reinitialize audio stream if its device isn't changed
### Step to reproduce

0. Make sure the computer has at least two input/output devices, and device S is the system default input/output device
1. Initialize an input/output stream with device X, where X is not the system default input device.
2. Change the system default input/output device to another device

### Expect
The audio stream keeps playing without being initialized.

### Actual
The audio stream is re-initialized even though nothing change for its device

ChunMinChang:
Currently we can get system-default-device-changed event even when we choose not to use the system default device, and every time when we get that event, we will reinitialize the audio stream. This makes the stream be reinitialized even if nothing changes on its device.
    
We should only reinitialize the audio stream when something happens on its device. This PR fixes input part of #149. We don't monitor alive-status for output device for now, hence the patches here fixes only input part.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -750,27 +750,12 @@ extern "C" fn audiounit_property_listener_callback(
     }
 
     // Handle the events
-
     if input_device_dead {
-        if !stm
-            .core_stream_data
-            .input_device
-            .flags
-            .contains(device_flags::DEV_SELECTED_DEFAULT)
-        {
-            cubeb_log!("The user-selected input device is dead, enter error state");
-            stm.notify_state_changed(State::Error);
-            stm.switching_device.store(false, Ordering::SeqCst);
-            return NO_ERR;
-        } else if addrs.len() == 1 {
-            // If this is the default input device ignore the event,
-            // kAudioHardwarePropertyDefaultInputDevice will take care of the switch
-            cubeb_log!("It's the default input device, ignore the event");
-            stm.switching_device.store(false, Ordering::SeqCst);
-            return NO_ERR;
-        }
+        cubeb_log!("The user-selected input device is dead, enter error state");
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -2957,20 +2969,27 @@ impl<'ctx> CoreStreamData<'ctx> {
...
-                cubeb_log!("AudioObjectAddPropertyListener/input/kAudioDevicePropertyDeviceIsAlive rv={}, device id ={}", rv, self.input_device.id);
-                return Err(Error::error());
+            // Get the notification when the input device is going away
+            // if the input doesn't follow the system default.
+            if !self
+                .input_device
+                .flags
+                .contains(device_flags::DEV_SELECTED_DEFAULT)
+            {
+                self.input_alive_listener = Some(device_property_listener::new(
```
ChunMinChang, line 2910: We probably don't need this if output_alive_listener monitors the same device.

src/backend/mod.rs
```diff
@@ -2869,6 +2837,23 @@ impl<'ctx> CoreStreamData<'ctx> {
             return Err(r);
         }
 
+        // We will have either *put_alive_listener or default_*put_listener.
+        // That is, we cannot have them both at the same time.
+        if self.has_input() {
```
padenot: This needs to go in the `assert!`, right ?

padenot (approved):
Please try to split your patches a lot more, with clear (and long if needed) commit messages for each part.

src/backend/mod.rs
ChunMinChang: I plan to do input side only for now in this PR.

padenot (approved):

Issue #70, opened 2020-03-27
Have an easy way to turn on the `cubeb_log!`


ChunMinChang:
The `cubeb_log!` won't print anything now. We should have a way to turn on those logs so it can help debugging.

ChunMinChang:
https://github.com/ChunMinChang/cubeb-coreaudio-rs/commit/323550d21b8644731c108fc4e6ac46d514f9c235 could work in *Rust Nightly* with`c_variadic` feature. 

Another workaround https://github.com/ChunMinChang/cubeb-coreaudio-rs/commit/9af97621f7af5a90630404fd30d48759761cba08 is using `cc` to build a custom log function in *C*. 
Otherwise we need to wait until https://github.com/rust-lang/rust/issues/44930 is fixed.

mutexlox-signal:
IIUC https://github.com/mozilla/cubeb-rs/pull/98 should fix this

PR #153, opened 2022-03-02
Reclock audio input, when needed

padenot:
This won't compile until the changes are merged in https://github.com/mozilla/cubeb/pull/690 and https://github.com/mozilla/cubeb-rs/pull/62.

When using aggregate devices, macOS reclocks for us. When not using them, it may be that they are still in the same clock domain, in which case no reclocking is necessary. If that's not the case, we want to reclock.

### DIFF
--- a/src/backend/buffer_manager.rs
+++ b/src/backend/buffer_manager.rs
@@ -2,11 +2,11 @@ use std::fmt;
 use std::os::raw::c_void;
 use std::slice;
 
-use cubeb_backend::{SampleFormat, StreamParams};
+use cubeb_backend::SampleFormat;
 
 use super::ringbuf::RingBuffer;
 
-use self::LinearInputBuffer::*;
+use self::LinearBuffer::*;
 use self::RingBufferConsumer::*;
 use self::RingBufferProducer::*;
 
@@ -75,16 +75,16 @@ fn remix_or_drop_channels<T: Copy + std::ops::Add<Output = T>>(
     output_channels * frame_count
 }
 
-fn process_input<T: Copy + std::ops::Add<Output = T>>(
-    input_data: *mut c_void,
+fn process_data<T: Copy + std::ops::Add<Output = T>>(
+    data: *mut c_void,
     frame_count: usize,
     input_channel_count: usize,
     input_channels_to_ignore: usize,
-    final_input_channels: usize,
+    output_channel_count: usize,
 ) -> &'static [T] {
-    assert!(input_channel_count >= input_channels_to_ignore);
+    assert!(input_channel_count >= input_channels_to_ignore + output_channel_count); // Only support downmix.
     let input_slice = unsafe {
-        slice::from_raw_parts_mut::<T>(input_data as *mut T, frame_count * input_channel_count)
+        slice::from_raw_parts_mut::<T>(data as *mut T, frame_count * input_channel_count)
     };
     if input_channels_to_ignore == 0 {
         input_slice
@@ -97,11 +97,11 @@ fn process_input<T: Copy + std::ops::Add<Output = T>>(
         );
         let new_count_remixed = remix_or_drop_channels(
             input_channel_count - input_channels_to_ignore,
-            final_input_channels,
+            output_channel_count,
             input_slice,
             frame_count,
         );
-        unsafe { slice::from_raw_parts_mut::<T>(input_data as *mut T, new_count_remixed) }
+        unsafe { slice::from_raw_parts_mut::<T>(data as *mut T, new_count_remixed) }
     }
 }
 
@@ -115,79 +115,90 @@ pub enum RingBufferProducer {
     FloatRingBufferProducer(ringbuf::Producer<f32>),
 }
 
-pub enum LinearInputBuffer {
-    IntegerLinearInputBuffer(Vec<i16>),
-    FloatLinearInputBuffer(Vec<f32>),
+pub enum LinearBuffer {
+    IntegerLinearBuffer(Vec<i16>),
+    FloatLinearBuffer(Vec<f32>),
 }
 
 pub struct BufferManager {
     consumer: RingBufferConsumer,
     producer: RingBufferProducer,
-    linear_input_buffer: LinearInputBuffer,
-    // Number of input channels that will be appended to the ring buffer, i.e. the number of
-    // channels that will be presented to the user callback.
-    input_channels: usize,
-    // Number of input channels that needs to be ignored, coming in.
+    linear_buffer: LinearBuffer,
+    // The number of channels in the interleaved data given to push_data
+    input_channel_count: usize,
+    // The number of channels that needs to be skipped in the beginning of input_channel_count
     input_channels_to_ignore: usize,
+    // The number of channels we actually needs, which is also the channel count of the
+    // processed data stored in the internal ring buffer.
+    output_channel_count: usize,
 }
 
 impl BufferManager {
     pub fn new(
-        params: &StreamParams,
-        input_channels_to_ignore: u32,
-        input_buffer_size_frames: u32,
-    ) -> BufferManager {
-        let format = params.format();
-        let input_channels = params.channels() as usize;
+        format: SampleFormat,
+        buffer_size_frames: usize,
+        input_channel_count: usize,
+        input_channels_to_ignore: usize,
+        output_channel_count: usize,
+    ) -> Self {
+        assert!(input_channel_count >= input_channels_to_ignore + output_channel_count);
         // 8 times the expected callback size, to handle the input callback being caled multiple
         //   times in a row correctly.
-        let buffer_element_count = input_channels * input_buffer_size_frames as usize * 8;
-        if format == SampleFormat::S16LE || format == SampleFormat::S16BE {
-            let ring = RingBuffer::<i16>::new(buffer_element_count);
-            let (prod, cons) = ring.split();
-            BufferManager {
-                producer: IntegerRingBufferProducer(prod),
-                consumer: IntegerRingBufferConsumer(cons),
-                linear_input_buffer: IntegerLinearInputBuffer(Vec::<i16>::with_capacity(
-                    buffer_element_count,
-                )),
-                input_channels,
-                input_channels_to_ignore: input_channels_to_ignore as usize,
+        let buffer_element_count = output_channel_count * buffer_size_frames as usize * 8;
+        match format {
+            SampleFormat::S16LE | SampleFormat::S16BE | SampleFormat::S16NE => {
+                let ring = RingBuffer::<i16>::new(buffer_element_count);
+                let (prod, cons) = ring.split();
+                Self {
+                    producer: IntegerRingBufferProducer(prod),
+                    consumer: IntegerRingBufferConsumer(cons),
+                    linear_buffer: IntegerLinearBuffer(Vec::<i16>::with_capacity(
+                        buffer_element_count,
+                    )),
+                    input_channel_count,
+                    input_channels_to_ignore,
+                    output_channel_count,
+                }
             }
-        } else {
-            let ring = RingBuffer::<f32>::new(buffer_element_count);
-            let (prod, cons) = ring.split();
-            BufferManager {
-                producer: FloatRingBufferProducer(prod),
-                consumer: FloatRingBufferConsumer(cons),
-                linear_input_buffer: FloatLinearInputBuffer(Vec::<f32>::with_capacity(
-                    buffer_element_count,
-                )),
-                input_channels,
-                input_channels_to_ignore: input_channels_to_ignore as usize,
+            SampleFormat::Float32LE | SampleFormat::Float32BE | SampleFormat::Float32NE => {
+                let ring = RingBuffer::<f32>::new(buffer_element_count);
+                let (prod, cons) = ring.split();
+                Self {
+                    producer: FloatRingBufferProducer(prod),
+                    consumer: FloatRingBufferConsumer(cons),
+                    linear_buffer: FloatLinearBuffer(Vec::<f32>::with_capacity(
+                        buffer_element_count,
+                    )),
+                    input_channel_count,
+                    input_channels_to_ignore,
+                    output_channel_count,
+                }
             }
         }
     }
-    pub fn push_data(&mut self, input_data: *mut c_void, frame_count: usize, channel_count: usize) {
-        let to_push = frame_count * self.input_channels;
+    pub fn channel_count(&self) -> usize {
+        self.output_channel_count
+    }
+    pub fn push_data(&mut self, data: *mut c_void, frame_count: usize) {
+        let to_push = frame_count * self.output_channel_count;
         let pushed = match &mut self.producer {
             RingBufferProducer::FloatRingBufferProducer(p) => {
-                let processed_input = process_input(
-                    input_data,
+                let processed_input = process_data(
+                    data,
                     frame_count,
-                    channel_count,
+                    self.input_channel_count,
                     self.input_channels_to_ignore,
-                    self.input_channels,
+                    self.output_channel_count,
                 );
                 p.push_slice(processed_input)
             }
             RingBufferProducer::IntegerRingBufferProducer(p) => {
-                let processed_input = process_input(
-                    input_data,
+                let processed_input = process_data(
+                    data,
                     frame_count,
-                    channel_count,
+                    self.input_channel_count,
                     self.input_channels_to_ignore,
-                    self.input_channels,
+                    self.output_channel_count,
                 );
                 p.push_slice(processed_input)
             }
@@ -200,12 +211,11 @@ impl BufferManager {
             );
         }
     }
-    fn pull_data(&mut self, input_data: *mut c_void, needed_samples: usize) {
+    fn pull_data(&mut self, data: *mut c_void, needed_samples: usize) {
         match &mut self.consumer {
             IntegerRingBufferConsumer(p) => {
-                let input: &mut [i16] = unsafe {
-                    slice::from_raw_parts_mut::<i16>(input_data as *mut i16, needed_samples)
-                };
+                let input: &mut [i16] =
+                    unsafe { slice::from_raw_parts_mut::<i16>(data as *mut i16, needed_samples) };
                 let read = p.pop_slice(input);
                 if read < needed_samples {
                     for i in 0..(needed_samples - read) {
@@ -214,9 +224,8 @@ impl BufferManager {
                 }
             }
             FloatRingBufferConsumer(p) => {
-                let input: &mut [f32] = unsafe {
-                    slice::from_raw_parts_mut::<f32>(input_data as *mut f32, needed_samples)
-                };
+                let input: &mut [f32] =
+                    unsafe { slice::from_raw_parts_mut::<f32>(data as *mut f32, needed_samples) };
                 let read = p.pop_slice(input);
                 if read < needed_samples {
                     for i in 0..(needed_samples - read) {
@@ -227,12 +236,12 @@ impl BufferManager {
         }
     }
     pub fn get_linear_data(&mut self, nsamples: usize) -> *mut c_void {
-        let p = match &mut self.linear_input_buffer {
-            LinearInputBuffer::IntegerLinearInputBuffer(b) => {
+        let p = match &mut self.linear_buffer {
+            LinearBuffer::IntegerLinearBuffer(b) => {
                 b.resize(nsamples, 0);
                 b.as_mut_ptr() as *mut c_void
             }
-            LinearInputBuffer::FloatLinearInputBuffer(b) => {
+            LinearBuffer::FloatLinearBuffer(b) => {
                 b.resize(nsamples, 0.);
                 b.as_mut_ptr() as *mut c_void
             }
--- a/src/backend/buffer_manager.rs
+++ b/src/backend/buffer_manager.rs
@@ -37,8 +37,9 @@ impl BufferManager {
     pub fn new(params: &StreamParams, input_buffer_size_frames: u32) -> BufferManager {
         let format = params.format();
         let channel_count = params.channels();
-        // Multiply by two for good measure.
-        let buffer_element_count = (channel_count * input_buffer_size_frames * 2) as usize;
+        // 8 times the expected callback size, to handle the input callback being caled multiple
+        //   times in a row correctly.
+        let buffer_element_count = (channel_count * input_buffer_size_frames * 8) as usize;
         if format == SampleFormat::S16LE || format == SampleFormat::S16BE {
             let ring = RingBuffer::<i16>::new(buffer_element_count);
             let (prod, cons) = ring.split();
--- a/src/backend/buffer_manager.rs
+++ b/src/backend/buffer_manager.rs
@@ -10,6 +10,100 @@ use self::LinearInputBuffer::*;
 use self::RingBufferConsumer::*;
 use self::RingBufferProducer::*;
 
+// Shuffles the data so that the first n channels of the interleaved buffer are overwritten by
+// the remaining channels.
+fn drop_first_n_channels_in_place<T: Copy>(
+    n: usize,
+    data: &mut [T],
+    frame_count: usize,
+    channel_count: usize,
+) {
+    // This function works if the numbers are equal but it's not particularly useful, so we hope to
+    // catch issues by checking using > and not >= here.
+    assert!(channel_count > n);
+    let mut read_idx: usize = 0;
+    let mut write_idx: usize = 0;
+
+    let channel_to_keep = channel_count - n;
+    for _ in 0..frame_count {
+        read_idx += n;
+        for _ in 0..channel_to_keep {
+            data[write_idx] = data[read_idx];
+            read_idx += 1;
+            write_idx += 1;
+        }
+    }
+}
+
+// It can be that the a stereo microphone is in use, but the user asked for mono input. In this
+// particular case, downmix the stereo pair into a mono channel. In all other cases, simply drop
+// the remaining channels before appending to the ringbuffer, becauses there is no right or wrong
+// way to do this, unlike with the output side, where proper channel matrixing can be done.
+// Return the number of valid samples in the buffer.
+fn remix_or_drop_channels<T: Copy + std::ops::Add<Output = T>>(
+    input_channels: usize,
+    output_channels: usize,
+    data: &mut [T],
+    frame_count: usize,
+) -> usize {
+    assert!(input_channels >= output_channels);
+    // Nothing to do, just return
+    if input_channels == output_channels {
+        return output_channels * frame_count;
+    }
+    // Simple stereo downmix
+    if input_channels == 2 && output_channels == 1 {
+        let mut read_idx = 0;
+        for (write_idx, _) in (0..frame_count).enumerate() {
+            data[write_idx] = data[read_idx] + data[read_idx + 1];
+            read_idx += 2;
+        }
+        return output_channels * frame_count;
+    }
+    // Drop excess channels
+    let mut read_idx = 0;
+    let mut write_idx = 0;
+    let channel_dropped_count = input_channels - output_channels;
+    for _ in 0..frame_count {
+        for _ in 0..output_channels {
+            data[write_idx] = data[read_idx];
+            write_idx += 1;
+            read_idx += 1;
+        }
+        read_idx += channel_dropped_count;
+    }
+    output_channels * frame_count
+}
+
+fn process_input<T: Copy + std::ops::Add<Output = T>>(
+    input_data: *mut c_void,
+    frame_count: usize,
+    input_channel_count: usize,
+    input_channels_needed: usize,
+) -> &'static [T] {
+    assert!(input_channel_count >= input_channels_needed);
+    let input_slice = unsafe {
+        slice::from_raw_parts_mut::<T>(input_data as *mut T, frame_count * input_channel_count)
+    };
+    if input_channel_count == input_channels_needed {
+        input_slice
+    } else {
+        drop_first_n_channels_in_place(
+            input_channel_count - input_channels_needed,
+            input_slice,
+            frame_count,
+            input_channel_count,
+        );
+        let new_count_remixed = remix_or_drop_channels(
+            input_channel_count,
+            input_channels_needed,
+            input_slice,
+            frame_count,
+        );
+        unsafe { slice::from_raw_parts_mut::<T>(input_data as *mut T, new_count_remixed) }
+    }
+}
+
 pub enum RingBufferConsumer {
     IntegerRingBufferConsumer(ringbuf::Consumer<i16>),
     FloatRingBufferConsumer(ringbuf::Consumer<f32>),
@@ -29,17 +123,16 @@ pub struct BufferManager {
     consumer: RingBufferConsumer,
     producer: RingBufferProducer,
     linear_input_buffer: LinearInputBuffer,
+    input_channels: usize,
 }
 
 impl BufferManager {
-    // When opening a duplex stream, the sample-spec are guaranteed to match. It's ok to have
-    // either the input or output sample-spec here.
     pub fn new(params: &StreamParams, input_buffer_size_frames: u32) -> BufferManager {
         let format = params.format();
-        let channel_count = params.channels();
+        let input_channels = params.channels() as usize;
         // 8 times the expected callback size, to handle the input callback being caled multiple
         //   times in a row correctly.
-        let buffer_element_count = (channel_count * input_buffer_size_frames * 8) as usize;
+        let buffer_element_count = input_channels * input_buffer_size_frames as usize * 8;
         if format == SampleFormat::S16LE || format == SampleFormat::S16BE {
             let ring = RingBuffer::<i16>::new(buffer_element_count);
             let (prod, cons) = ring.split();
@@ -49,6 +142,7 @@ impl BufferManager {
                 linear_input_buffer: IntegerLinearInputBuffer(Vec::<i16>::with_capacity(
                     buffer_element_count,
                 )),
+                input_channels,
             }
         } else {
             let ring = RingBuffer::<f32>::new(buffer_element_count);
@@ -59,27 +153,29 @@ impl BufferManager {
                 linear_input_buffer: FloatLinearInputBuffer(Vec::<f32>::with_capacity(
                     buffer_element_count,
                 )),
+                input_channels,
             }
         }
     }
-    pub fn push_data(&mut self, input_data: *const c_void, read_samples: usize) {
+    pub fn push_data(&mut self, input_data: *mut c_void, frame_count: usize, channel_count: usize) {
+        let to_push = frame_count * self.input_channels;
         let pushed = match &mut self.producer {
             RingBufferProducer::FloatRingBufferProducer(p) => {
-                let input_data =
-                    unsafe { slice::from_raw_parts::<f32>(input_data as *const f32, read_samples) };
-                p.push_slice(input_data)
+                let processed_input =
+                    process_input(input_data, frame_count, channel_count, self.input_channels);
+                p.push_slice(processed_input)
             }
             RingBufferProducer::IntegerRingBufferProducer(p) => {
-                let input_data =
-                    unsafe { slice::from_raw_parts::<i16>(input_data as *const i16, read_samples) };
-                p.push_slice(input_data)
+                let processed_input =
+                    process_input(input_data, frame_count, channel_count, self.input_channels);
+                p.push_slice(processed_input)
             }
         };
-        if pushed != read_samples {
+        if pushed != to_push {
             cubeb_log!(
                 "Input ringbuffer full, could only push {} instead of {}",
                 pushed,
-                read_samples
+                to_push
             );
         }
     }
--- a/src/backend/buffer_manager.rs
+++ b/src/backend/buffer_manager.rs
@@ -79,24 +79,25 @@ fn process_input<T: Copy + std::ops::Add<Output = T>>(
     input_data: *mut c_void,
     frame_count: usize,
     input_channel_count: usize,
-    input_channels_needed: usize,
+    input_channels_to_ignore: usize,
+    final_input_channels: usize,
 ) -> &'static [T] {
-    assert!(input_channel_count >= input_channels_needed);
+    assert!(input_channel_count >= input_channels_to_ignore);
     let input_slice = unsafe {
         slice::from_raw_parts_mut::<T>(input_data as *mut T, frame_count * input_channel_count)
     };
-    if input_channel_count == input_channels_needed {
+    if input_channels_to_ignore == 0 {
         input_slice
     } else {
         drop_first_n_channels_in_place(
-            input_channel_count - input_channels_needed,
+            input_channels_to_ignore,
             input_slice,
             frame_count,
             input_channel_count,
         );
         let new_count_remixed = remix_or_drop_channels(
-            input_channel_count,
-            input_channels_needed,
+            input_channel_count - input_channels_to_ignore,
+            final_input_channels,
             input_slice,
             frame_count,
         );
@@ -123,11 +124,19 @@ pub struct BufferManager {
     consumer: RingBufferConsumer,
     producer: RingBufferProducer,
     linear_input_buffer: LinearInputBuffer,
+    // Number of input channels that will be appended to the ring buffer, i.e. the number of
+    // channels that will be presented to the user callback.
     input_channels: usize,
+    // Number of input channels that needs to be ignored, coming in.
+    input_channels_to_ignore: usize,
 }
 
 impl BufferManager {
-    pub fn new(params: &StreamParams, input_buffer_size_frames: u32) -> BufferManager {
+    pub fn new(
+        params: &StreamParams,
+        input_channels_to_ignore: u32,
+        input_buffer_size_frames: u32,
+    ) -> BufferManager {
         let format = params.format();
         let input_channels = params.channels() as usize;
         // 8 times the expected callback size, to handle the input callback being caled multiple
@@ -143,6 +152,7 @@ impl BufferManager {
                     buffer_element_count,
                 )),
                 input_channels,
+                input_channels_to_ignore: input_channels_to_ignore as usize,
             }
         } else {
             let ring = RingBuffer::<f32>::new(buffer_element_count);
@@ -154,6 +164,7 @@ impl BufferManager {
                     buffer_element_count,
                 )),
                 input_channels,
+                input_channels_to_ignore: input_channels_to_ignore as usize,
             }
         }
     }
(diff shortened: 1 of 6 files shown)


src/backend/mod.rs
```diff
@@ -2427,6 +2427,23 @@ impl<'ctx> CoreStreamData<'ctx> {
             && self.input_device.id != self.output_device.id
             && !is_device_a_type_of(self.input_device.id, DeviceType::OUTPUT)
             && !is_device_a_type_of(self.output_device.id, DeviceType::INPUT)
+            && false
```
ChunMinChang: This makes `should_use_aggregate_device` always returns `false`. Does it add for debugging only and forget to clean? or is there a reason to avoid using aggregate device?

src/backend/mod.rs
```diff
@@ -2427,6 +2427,23 @@ impl<'ctx> CoreStreamData<'ctx> {
...
+
+    fn same_clock_domain(&self) -> bool {
+        // If not setting up a duplex stream, there is only one device,
+        // no reclocking necessary.
+        if !(self.has_input() && self.has_output()) {
+            return true;
+        }
+        if let Ok(input_domain) = get_clock_domain(self.input_device.id, DeviceType::INPUT) {
+            if let Ok(output_domain) = get_clock_domain(self.output_device.id, DeviceType::OUTPUT) {
+                println!("Clock domains {} {}", input_domain, output_domain);
```
ChunMinChang: Replace `println!` by `cubeb_log!`.

I guess this is written when you're debugging. This reminds me https://github.com/mozilla/cubeb-coreaudio-rs/issues/70. We should have an easier way to enable the `cubeb_log` here.

src/backend/mod.rs
```diff
@@ -2427,6 +2427,23 @@ impl<'ctx> CoreStreamData<'ctx> {
...
+                return input_domain == output_domain;
+            }
+        }
+        cubeb_log!("Coudn't determine clock domains for input and output");
```
ChunMinChang: `cubeb_log!("Coudn't determine clock domains for input or output")`?

Or, to have a better error messages, why don't we use
```rs
let input_domain = match get_clock_domain(self.input_device.id, DeviceType::INPUT) {
    Ok(domain) => domain,
    Err(e) => {
        cubeb_log!("Fail to get clock domains for input. Error: {}", e);
        return false;
    }
}

let output_domain = ... // Do the same

input_domain == output_domain
```

src/backend/mod.rs
```diff
@@ -2446,6 +2463,8 @@ impl<'ctx> CoreStreamData<'ctx> {
 
         let mut in_dev_info = self.input_device.clone();
         let mut out_dev_info = self.output_device.clone();
+        let mut using_aggregate_device = false;
```
ChunMinChang: Instead of adding a `using_aggregate_device` here, we should make `aggregate_device: Option<AggregateDevice>`. That would be much easier to know if we have an aggregate_device in use or not, and worry-free to synchronize `aggregate_device` with `using_aggregate_device: bool`.

ChunMinChang (changes requested):
The idea of the logic here seems good, but I don't have the knowledge of the new `cubeb_resampler` design we are going to use. Could you give me some context?

Now https://github.com/mozilla/cubeb/pull/690 and https://github.com/mozilla/cubeb-rs/pull/62 are already merged, we should be able to make the tests work.

src/backend/mod.rs
padenot: A mistake.

src/backend/mod.rs
padenot: I'm removing this anyways.

src/backend/mod.rs
padenot: much better style, thanks.

src/backend/mod.rs
padenot: Yes, better. But it will always be true now. I'll like to keep it like that (=have a function to enable/disable it, instead of cleaning everything up) for now, until we test more. There might be some circumstances where we don't want aggregate devices.

padenot: But we'll only know those possible circumstances after testing more with lots of devices (I've tested a lot already, but I only have so many devices).

padenot:
> The idea of the logic here seems good, but I don't have the knowledge of the new `cubeb_resampler` design we are going to use. Could you give me some context?
> 
> Now [mozilla/cubeb#690](https://github.com/mozilla/cubeb/pull/690) and [mozilla/cubeb-rs#62](https://github.com/mozilla/cubeb-rs/pull/62) are already merged, we should be able to make the tests work.

Generally, if the resampler finds that there are not enough input frames compared to output frames, it will very carefully strech the input audio.

Here is a practical example. If I connect a Microsoft Lifecam VX5000 to my Macbook Pro 2018, and use it as a microphone, and then open a duplex stream with the built-in headphones for the output, with the input and output at the same rate, I receive (on average, after running a long time), 0.04% more audio input frames that there are output frames rendered.

So my new system will detect this, and adjust a resampler for the input stream by a minuscule amount so that no extra buffering happens. If the opposite hapens (input doesn't deliver enough frames), it will stretch the audio.

But with the last patches in this PR, aggregate devices always work, and macOS reclocks for us, so this doesn't happen anymore. My resampling patches will be useful for Windows though.

src/backend/mod.rs
```diff
@@ -667,6 +669,7 @@ extern "C" fn audiounit_output_callback(
         stm.stopped.store(true, Ordering::SeqCst);
         stm.core_stream_data.stop_audiounits();
         audiounit_make_silent(&mut buffers[0]);
+        cubeb_log!("Error callback: outframes is {}", outframes);
```
ChunMinChang: The commit message says "Logging statements that are in real-time threads should use cubeb_logv" but `cubeb_logv!` is applied here. Shouldn't this be `cubeb_logv!`?

src/backend/mod.rs
```diff
@@ -2267,8 +2271,11 @@ unsafe impl Sync for AudioUnitContext {}
 #[derive(Debug)]
 struct CoreStreamData<'ctx> {
     stm_ptr: *const AudioUnitStream<'ctx>,
-    aggregate_device: AggregateDevice,
+    aggregate_device: Option<AggregateDevice>,
```
ChunMinChang, line 2197: Oh this looks much better! Thanks!

src/backend/mod.rs
```diff
@@ -2523,6 +2559,11 @@ impl<'ctx> CoreStreamData<'ctx> {
             );
             self.input_hw_rate = input_hw_desc.mSampleRate;
 
+            // Can't open a stream with more input channels than the device has
+            if input_hw_desc.mChannelsPerFrame < self.input_stream_params.channels() {
```
ChunMinChang: Is `input_hw_desc.mChannelsPerFrame` the max input channel count of the device?

src/backend/mod.rs
```diff
@@ -2267,8 +2271,11 @@ unsafe impl Sync for AudioUnitContext {}
 #[derive(Debug)]
 struct CoreStreamData<'ctx> {
     stm_ptr: *const AudioUnitStream<'ctx>,
-    aggregate_device: AggregateDevice,
+    aggregate_device: Option<AggregateDevice>,
     mixer: Option<Mixer>,
+    // The input channels to ignore. They are always first in the list of channels, because of the
+    // way the aggregate device is setup.
+    input_channels_to_ignore: u32,
```
ChunMinChang: > because of the way the aggregate device is setup.

Can we change how we set up the aggregate device so we don't need this?

I am not sure I understand why we need this. Is input channel count for an aggregate device the total of input channel count of all the devices in aggregate device?

If we only use aggregate device when the output device has no input and the input device has no output, do we still need this?


If this is unavoidable, it should be moved into `AggregateDevice`, since it is only used when stream is using `aggregate_device`

src/backend/mod.rs
```diff
@@ -2532,6 +2573,19 @@ impl<'ctx> CoreStreamData<'ctx> {
...
+            // use here.
+            if self.input_channels_to_ignore != 0 {
+                if let Ok(count) = get_channel_count(in_dev_info.id, DeviceType::INPUT) {
+                    self.input_desc.mChannelsPerFrame = count;
+                    self.input_desc.mBytesPerFrame =
+                        (self.input_desc.mBitsPerChannel / 8) * self.input_desc.mChannelsPerFrame;
+                    self.input_desc.mBytesPerPacket =
+                        self.input_desc.mBytesPerFrame * self.input_desc.mFramesPerPacket;
+                }
+            }
```
ChunMinChang: Will `get_channel_count(in_dev_info.id, DeviceType::INPUT)` return sum of all the available input count of the devices listed in the aggregate device?

I find the above code duplicates in several places. Can you write a util function for them?

src/backend/mod.rs
```diff
@@ -2414,18 +2423,32 @@ impl<'ctx> CoreStreamData<'ctx> {
...
-        // microphone of the aggregate device for this duplex stream.
-        self.has_input()
-            && self.has_output()
-            && self.input_device.id != kAudioObjectUnknown
-            && self.input_device.flags.contains(device_flags::DEV_INPUT)
-            && self.output_device.id != kAudioObjectUnknown
-            && self.output_device.flags.contains(device_flags::DEV_OUTPUT)
-            && self.input_device.id != self.output_device.id
-            && !is_device_a_type_of(self.input_device.id, DeviceType::OUTPUT)
-            && !is_device_a_type_of(self.output_device.id, DeviceType::INPUT)
```
ChunMinChang, line 2363: This seems to cause regression for #13 and [BMO 1563475](https://bugzilla.mozilla.org/show_bug.cgi?id=1563475)

src/backend/buffer_manager.rs
```diff
@@ -10,6 +10,94 @@ use self::LinearInputBuffer::*;
...
+    let channel_to_keep = channel_count - n;
+    for _ in 0..frame_count {
+        read_idx += n;
+        for _ in 0..channel_to_keep {
+            data[write_idx] = data[read_idx];
+            read_idx += 1;
+            write_idx += 1;
+        }
+    }
+}
```
ChunMinChang, line 36: Nice!

src/backend/buffer_manager.rs
```diff
@@ -10,6 +10,94 @@ use self::LinearInputBuffer::*;
...
+    }
+    output_channels * frame_count
+}
+
+fn process_input<T: Copy + std::ops::Add<Output = T>>(
+    input_data: *mut c_void,
+    frame_count: usize,
+    channel_count: usize,
+    output_channels: usize,
+    input_channels_to_ignore: usize,
```
ChunMinChang, line 82: What's the relationship between `channel_count`, `output_channels`, and `input_channels_to_ignore`?
`channel_count >= input_channels_to_ignore + output_channels`? Add an `assert!`?

src/backend/buffer_manager.rs
```diff
@@ -58,27 +149,45 @@ impl BufferManager {
...
         }
     }
-    pub fn push_data(&mut self, input_data: *const c_void, read_samples: usize) {
+    pub fn push_data(
+        &mut self,
+        input_data: *mut c_void,
+        frame_count: usize,
+        channel_count: usize,
+        input_channels_to_ignore: usize,
+    ) {
```
ChunMinChang: `assert!(channel_count >= input_channels_to_ignore + self.output_channels)`?

src/backend/mod.rs
```diff
@@ -412,10 +412,12 @@ extern "C" fn audiounit_input_callback(
...
         } else {
             assert_eq!(status, NO_ERR);
-            // Copy input data in linear buffer.
-            let elements =
-                (input_frames * stm.core_stream_data.input_desc.mChannelsPerFrame) as usize;
-            input_buffer_manager.push_data(input_buffer_list.mBuffers[0].mData, elements);
+            input_buffer_manager.push_data(
+                input_buffer_list.mBuffers[0].mData,
+                input_frames as usize,
+                stm.core_stream_data.input_desc.mChannelsPerFrame as usize,
```
ChunMinChang: Now `stm.core_stream_data.input_desc.mChannelsPerFrame` is the max hardware input channel count?

src/backend/mod.rs
```diff
@@ -1308,6 +1311,7 @@ fn audiounit_get_default_datasource_string(devtype: DeviceType) -> Result<CStrin
     Ok(convert_uint32_into_string(data))
 }
 
+#[cfg(test)]
```
ChunMinChang: This could be removed if no other non-test code needs it.

src/backend/mod.rs
```diff
@@ -2532,6 +2573,19 @@ impl<'ctx> CoreStreamData<'ctx> {
                     );
                     e
                 })?;
+            // We're using an aggregate device with the output device having input channels.
+            // It's necessary to ignore those input channels, and then to remix the remaining
+            // input channels into the format requested. In anycase, all channels should be in
+            // use here.
+            if self.input_channels_to_ignore != 0 {
```
ChunMinChang: if self.aggregate_device.is_some()

src/backend/tests/manual.rs
```diff
@@ -1,6 +1,6 @@
 use super::utils::{
     test_get_devices_in_scope, test_ops_context_operation, test_ops_stream_operation, Scope,
-    StreamType, TestDeviceSwitcher,
+    StreamType, TestDeviceInfo, TestDeviceSwitcher,
```
ChunMinChang: These code has been merged in trailblazer branch. Can you rebase your code?

src/backend/tests/device_change.rs
```diff
@@ -597,13 +597,17 @@ fn test_unplug_a_device_on_an_active_stream(
...
-                device_scope,
-                device_scope,
-                test_get_default_device(device_scope.clone()).unwrap()
-            );
+            if let Some(id) = test_get_default_device(device_scope.clone()) {
+                println!(
+                    "Device {} for {:?} is unplugged. The default {:?} device now is {}",
+                    dev, device_scope, device_scope, id
+                );
+            } else {
```
ChunMinChang: I don't think it's possible here. Are you running into this problem?

The `TestDevicePlugger` is implemented by an aggregate device, and it will be created when we have at least one input or output device. When the `plugger.unplug()` hads been executed, the new default device should be set to the original default device.

ChunMinChang (changes requested):
It seems these changes will always use hardware input channel count to get the input data, from all the (sub-)devices listed in the aggregate device, and then downmix them if necessary.

Can you describe a way so I can test this patch?

And, some commit order causes build error (e.g., `Make aggregate_device an Option` use `same_clock_domain` before it's implemented ) so we can only squash these commits before merging it. If you'd like "rebase and merge", please reorder them so it's easier to revert some of them if we are unlucky.

src/backend/mod.rs
padenot: Yeah there are two problems.

First, I thought that `cubeb_logv!` was using the [async logging facility](https://searchfox.org/mozilla-central/source/media/libcubeb/src/cubeb_log.h#55-60), but that's not the case, and it's a big issue, because it means that logging isn't real-time safe. If I add a couple of logging statements to a debug build, and run with low-latency on my macbook 2018, it glitches...

So I'm going to send a patch to add it to `cubeb-rs`, but that's going to take a long time. And then I'm going to fix it here, so we'll be able to have proper logging from real-time threads.

Second, my message isn't correct. I conflated the messages that are frequent, and the message that are logged from the real-time thread, which are often the same messages, but not always.

Messages that aren't frequent, but from the real-time callback, should use normal async logging (non-verbose). But async logging isn't exposed in cubeb-rs so this is normal logging _for now_.

I filed https://github.com/mozilla/cubeb-rs/issues/66 for this and I'll drop this commit. Those message should go in normal log I was wrong.

src/backend/mod.rs
padenot: No, we can't change this, it's how it works: an aggregate device puts two devices together as a single device. It's useful on its own. So if you aggregate a sound card that has 4 output and 4 input with a Bluetooth headset that has 2 output and 1 input, the resulting device has 6 output, 5 inputs, that's it.

If we aggregate a device that has just 2 channels of output with a device that has 1 channel of input, this number `input_chanenls_to_ignore` 0, and everything works well, no reordering will happen.

IMO this is not a something the aggregate device should do the `Aggregatedevice` class handles creation and sub-devices, nothing else, and especially nothing during processing: we don't want to special case aggregate device and regular devices during processing.

This is something that this cubeb backend wants to do: it wants to only ever use some specific channels in an aggregate device to implement a duplex stream with reclocking.

src/backend/mod.rs
padenot: > Will get_channel_count(in_dev_info.id, DeviceType::INPUT) return sum of all the available input count of the devices listed in the aggregate device?

Yes, if an aggregate device has two sub-devices, one with 2 channels of input one with 1 channel of input, calling this with the aggregate device ID will return 3.

I've added a commit turn a function for this.

src/backend/mod.rs
padenot: No, that's the whole point of this patch set, being able to use an aggregate device (and its benefits, mainly drift compensation) and using the intended device.

src/backend/buffer_manager.rs
padenot: It's confusing, I'll rename and assert.

src/backend/mod.rs
padenot: Yes, always.

src/backend/mod.rs
padenot: It's needed, I tried to remove it, but it broke `cargo test`.

src/backend/tests/device_change.rs
padenot: Yes, devices disappear from the test machine. Subsequently, trying to open a duplex stream fails (the state callback is called with `Error`). Locally it works well, it seems like.

src/backend/mod.rs
padenot: No. We can have an aggregate device and not ignore any channels, when aggregating with an output device that only has output channels (not input channels).

padenot:
> Can you describe a way so I can test this patch?

Find a device (A) that has both input and output channels. Examples: the Blackhole virtual device, most pro audio soundcard that have input and output jacks to connect musical instruments. If using the Blackhole virtual device you'll want to listen to the output side via something else, e.g. a DAW. I'm using either Blackhole or an old Focusrite Scarlett 6i6.

Find a device (B) that has only an input channel, such as the built-in mic of the macbook.

Set the default output device to (A), the default input device to (B), run any cubeb example that pipes the input to the output, and you should hear the correct input device in the correct output, but with an aggregate device (so reclocking is done for us internally).

src/backend/mod.rs
ChunMinChang: How do we know the `input_channels_to_ignore`, which is the input channel of the output-side device in aggregate device, always present in the first, before the input channel of the input-side device in aggregate device?

If we don't put `input_channels_to_ignore` in `Aggregatedevice`, can we put it in `input_buffer_manager` when  initializing it? It's a constant once we set `aggregate_device`

padenot: It's explained in the commit message: https://github.com/mozilla/cubeb-coreaudio-rs/pull/153/commits/fb8aeb99ba171b3ea4512266bb76c95a1888e0c8, and also I've done the change you suggested already.

src/backend/mod.rs
ChunMinChang: Sorry, I changed my mind.

Can use `get_channel_count(in_dev_info.id, DeviceType::INPUT)` as channel count and `self.input_hw_rate` as sample rate to create a temporary `StreamParams`, then use it to create `self.input_desc`? So we will end up having a `self.input_desc` with correct information we what and we can use it every where (`self.input_hw_rate` can probably go away).

src/backend/mod.rs
ChunMinChang: I mean the `is_device_a_type_of` and its test can be removed since no production code needs it. Does it still break `cargo test`?

src/backend/mod.rs
ChunMinChang: https://github.com/mozilla/cubeb-coreaudio-rs/blob/ca183a83c7c2d01749b97b26979bbd715b95dcd3/src/backend/aggregate_device.rs#L400-L401

I guess you mean here?

Just to make sure I understand. If we add input sub devices before output sub devices, then the first couple of input channel will be from the input sub devices, and we won't need `input_channels_to_ignore`, but we will end up having same situation in output side and need a  `output_channels_to_ignore`?

padenot: We'll still need `input_channels_to_ignore`, but the channel to ignore will be last, not first, in the buffers if the input device is first in the list.

src/backend/mod.rs
padenot: Not sure I understand what you mean by correct. How do you modify a `StreamParamRef` anyways ?

src/backend/mod.rs
ChunMinChang: > if you aggregate a sound card that has 4 output and 4 input with a Bluetooth headset that has 2 output and 1 input, the resulting device has 6 output, 5 inputs, that's it.

It sounds like we need to ignore the extra channel count on output side as well?

ChunMinChang: > It's explained in the commit message: https://github.com/mozilla/cubeb-coreaudio-rs/commit/fb8aeb99ba171b3ea4512266bb76c95a1888e0c8, and also I've done the change you suggested already.

The `input_channels_to_ignore` is unnecessary now by the change, right?

src/backend/mod.rs
ChunMinChang: I think we can do something like this

```rs
let device_input_channels = get_channel_count(in_dev_info.id, DeviceType::INPUT)
     .unwrap_or(self.input_stream_params.channels());

// Set format description according to the input params.
let params = unsafe {
    p.channels = device_input_channels;
    p.rate = input_hw_desc.mSampleRate as _;
    StreamParams::from(p)
 };

self.input_desc = create_stream_description(&params).map_err(|e| {
     cubeb_log!("({:p}) Setting format description for input failed.", self.stm_ptr);
     e
})?;

assert_eq!(self.input_desc.mSampleRate, input_hw_desc.mSampleRate);
assert_eq!(self.input_desc.mChannelsPerFrame, device_input_channels);
```

So `self.input_desc` will have the hardware information, and we can remove the `input_hw_rate` as well.

src/backend/mod.rs
ChunMinChang: Suppose aggregate device _A_ contains device _X_ and device _Y_. Device _X_ has _m_ input channels and _n_ output channels. Device _Y_ has _p_ input channels and _q_ output channels. I think my question is:

1. If we add _X_ before _Y_ in aggregate device _A_, what are the input channels and output channels of _A_ when we set input, output channel count to _i_ and _j_  by `audio_unit_set_property(..., kAudioUnitProperty_StreamFormat, ...)`, where _0 < i <= m + p_, and _0 < j <= n + q_?
2. Similar as above, if we add _Y_ before _X_ in aggregate device _A_, what will we get?

padenot: 1. We don't do that. We just grab all channels and figure out what channels we want to use. We always get all channels of input and only ever give the channels of the cubeb input device to the callback. For the output, we use the cubeb mixer to only use the channels of the cubueb output device (the others are `Silent`). 
2. We don't do that, but if we did, we'd get precisely the opposite ordering: we'd have to only take the first channels of the input buffers and discard the other channels (instead of taking the last), and we'd have to ask the cubeb mixer to put `Silent` to the channels corresponding to the output channels of the cubeb input device (which would be first), and only use the few last channels.

padenot: Actually I was wrong, we still need `input_channels_to_ignore` or something. We can have the following:

- Output device is 16in, 16out. Input device is 2in
- The user requests mono input
- `input_channels_to_ignore` is 16, number of channels total is 18, so there are 2 remaining channels. Now we need to down-mix those two channels to mono.

I've refactored the code because it's only needed at construction though, no need to store it in `CoreStreamData`.

src/backend/mod.rs
padenot: Yeah ok. Not sure that's better though, having an unsafe block in the middle of the code is not amazing.

src/backend/mod.rs
```diff
@@ -2500,17 +2518,36 @@ impl<'ctx> CoreStreamData<'ctx> {
...
+            let params = unsafe {
+                let mut p = *self.input_stream_params.as_ptr();
+                p.channels = device_input_channels;
+                // Input AudioUnit must be configured with device's sample rate.
+                // we will resample inside input callback.
+                p.rate = input_hw_desc.mSampleRate as _;
+                StreamParams::from(p)
+            };
+
+            self.input_desc = create_stream_description(&params).map_err(|e| {
```
ChunMinChang: I find we can do the similar thing for `output_desc`. Ideally, at the end we will merge `input_hw_rate` and `output_hw_rate` into `input_desc` and `output_desc`. And perhaps we should rename them to `input_dev_desc` and `output_dev_desc`

```rs
struct CoreStreamData<'ctx> {
    ...
    // Stream creation parameters.
    input_stream_params: StreamParams,
    output_stream_params: StreamParams,
    // Device settings for AudioUnits.
    input_dev_desc: AudioStreamBasicDescription,
    output_dev_desc: AudioStreamBasicDescription,
    ...
}
```

src/backend/mod.rs
```diff
@@ -2424,6 +2439,7 @@ impl<'ctx> CoreStreamData<'ctx> {
```
ChunMinChang, line 2409:
```suggestion
    fn get_device_channel_count(id: AudioDeviceID, devtype: DeviceType) -> Option<u32> {
            get_channel_count(id, devtype)
                .or_else(|e| {
                    cubeb_log!("Fail to get channel count for device {}. Error: {}", id, e);
                    get_device_stream_format(id, devtype).map(|f| f.mChannelsPerFrame)
                })
                .map_err(|e| {
                    cubeb_log!(
                        "Fail to get channel info from stream format for device {}. Error: {}",
                        id,
                        e
                    );
                })
                .ok()
        }
```
We can add a function that will be reused in both input side and output side, at the beginning of `setup`, to get the device channel count.

src/backend/mod.rs
ChunMinChang: I find this can be wrong when using aggregate device. `input_hw_desc.mChannelsPerFrame` here will be the total input channel count of all the sub-devices added in the aggregate device (e.g., we will get _m + n_ if the output device has _m_ input channels and the input device has _n_ input channels).

We should put something like this below `cubeb_log!("({:p}) Initialize input by device info: {:?}", ...)` and before `.. create_audiounit( ... ) ...` instead.

```rs
            let device_channel_count =
                get_device_channel_count(self.input_device.id, DeviceType::INPUT).unwrap_or(0);
            if device_channel_count < self.input_stream_params.channels() {
                return Err(Error::invalid_parameter());
            }
```

Ad we should also do the same thing for the output-side as well:
```rs
            let device_channel_count =
                get_device_channel_count(self.output_device.id, DeviceType::OUTPUT).unwrap_or(0);
            if device_channel_count < self.output_stream_params.channels() {
                return Err(Error::invalid_parameter());
            }
```

src/backend/mod.rs
```diff
@@ -2500,17 +2518,36 @@ impl<'ctx> CoreStreamData<'ctx> {
...
-                    );
-                    e
-                })?;
+            // Can't open a stream with more input channels than the device has
+            if input_hw_desc.mChannelsPerFrame < self.input_stream_params.channels() {
+                return Err(Error::error());
+            }
+
+            let device_input_channels = get_channel_count(in_dev_info.id, DeviceType::INPUT)
+                .unwrap_or_else(|_| self.input_stream_params.channels());
```
```suggestion
```
We don't need this if we call `get_device_channel_count` in the beginning for input settings.

src/backend/mod.rs
```diff
@@ -2500,17 +2518,36 @@ impl<'ctx> CoreStreamData<'ctx> {
...
+
+            // Always request all the input channels of the device, and only pass the correct
+            // channels to the audio callback.
+            let params = unsafe {
+                let mut p = *self.input_stream_params.as_ptr();
+                p.channels = device_input_channels;
```
```suggestion
                p.channels = input_hw_desc.mChannelsPerFrame;
```

`input_hw_desc.mChannelsPerFrame` is the sum of all the input channel count of the devices added in the aggregate device.

src/backend/mod.rs
```diff
@@ -2500,17 +2518,36 @@ impl<'ctx> CoreStreamData<'ctx> {
...
+            self.input_desc = create_stream_description(&params).map_err(|e| {
+                cubeb_log!(
+                    "({:p}) Setting format description for input failed.",
+                    self.stm_ptr
+                );
+                e
+            })?;
+
+            assert_eq!(self.input_desc.mSampleRate, input_hw_desc.mSampleRate);
+            assert_eq!(self.input_desc.mChannelsPerFrame, device_input_channels);
```
```suggestion
```

src/backend/mod.rs
```diff
@@ -2500,17 +2518,36 @@ impl<'ctx> CoreStreamData<'ctx> {
                 self.stm_ptr,
                 input_hw_desc
             );
-            self.input_hw_rate = input_hw_desc.mSampleRate;
```
```suggestion
assert!(input_hw_desc.mChannelsPerFrame >= device_channel_count);
// Notice: when we are using aggregate device, the input_hw_desc.mChannelsPerFrame is
// the total of all the input channel count of the devices added in the aggregate device.
// Due to our aggregate device settings, the data captured by the output device's input
// channels will be put in the beginning of the raw data given by the input callback.
```
We can add a line to make sure `input_hw_desc.mChannelsPerFrame` always have more or same number of channels we want.

src/backend/buffer_manager.rs
```diff
@@ -10,6 +10,101 @@ use self::LinearInputBuffer::*;
...
+    }
+    output_channels * frame_count
+}
+
+fn process_input<T: Copy + std::ops::Add<Output = T>>(
+    input_data: *mut c_void,
+    frame_count: usize,
+    input_channel_count: usize,
+    input_channels_to_ignore: usize,
+    final_input_channels: usize,
```
ChunMinChang:
```suggestion
fn process_data<T: Copy + std::ops::Add<Output = T>>(
    data: *mut c_void,
    frame_count: usize,
    input_channel_count: usize,
    input_channels_to_ignore: usize,
    output_channel_count: usize,
```

I'd like to unify the concept of `input` and `output` in this file, so I think we can rename this to `process_data` instead. See more in my comments below.

src/backend/buffer_manager.rs
```diff
@@ -29,16 +124,24 @@ pub struct BufferManager {
...
 
 impl BufferManager {
-    // When opening a duplex stream, the sample-spec are guaranteed to match. It's ok to have
-    // either the input or output sample-spec here.
-    pub fn new(params: &StreamParams, input_buffer_size_frames: u32) -> BufferManager {
+    pub fn new(
+        params: &StreamParams,
+        input_channels_to_ignore: u32,
+        input_buffer_size_frames: u32,
+    ) -> BufferManager {
```
ChunMinChang:
```suggestion
    format: SampleFormat,
        buffer_size_frames: usize,
        input_channel_count: usize,
        input_channels_to_ignore: usize,
        output_channel_count: usize,
    ) -> Self {
```

Since we need to do audio processing here, I'd like to match the concept of all the *input* and *output* in this file. The _input_channel*_ is the channel count of the data given to `push_data`, and the _output_channel_ is the channel count of the processed data we save in the internal ring buffer. Hence we should also rename the `LinearInputBuffer` to `LinearBuffer`

```rs
pub enum LinearBuffer {
    IntegerLinearBuffer(Vec<i16>),
    FloatLinearBuffer(Vec<f32>),
}
```

Does this make sense to you?

It would be better to put some comments somewhere as a note, in case we forget we we design `BufferManager` in this way.

```rs
    pub fn new(
        format: SampleFormat,
        buffer_size_frames: usize,
        input_channel_count: usize,
        input_channels_to_ignore: usize,
        output_channel_count: usize,
    ) -> Self {
        // We only support downmix in BufferManager, so output_channel_count must be equal to or less than the left channel count we can use.
        assert!(input_channel_count >= input_channels_to_ignore + output_channel_count);
        // 8 times the expected callback size, to handle the input callback being caled multiple
        //   times in a row correctly.
        let buffer_element_count = output_channel_count * buffer_size_frames as usize * 8;
        match format {
            SampleFormat::S16LE | SampleFormat::S16BE | SampleFormat::S16NE => {
                let ring = RingBuffer::<i16>::new(buffer_element_count);
                let (prod, cons) = ring.split();
                Self {
                    producer: IntegerRingBufferProducer(prod),
                    consumer: IntegerRingBufferConsumer(cons),
                    linear_buffer: IntegerLinearBuffer(Vec::<i16>::with_capacity(
                        buffer_element_count,
                    )),
                    input_channel_count,
                    input_channels_to_ignore,
                    output_channel_count,
                }
            }
            SampleFormat::Float32LE | SampleFormat::Float32BE | SampleFormat::Float32NE => {
                let ring = RingBuffer::<f32>::new(buffer_element_count);
                let (prod, cons) = ring.split();
                Self {
                    producer: FloatRingBufferProducer(prod),
                    consumer: FloatRingBufferConsumer(cons),
                    linear_buffer: FloatLinearBuffer(Vec::<f32>::with_capacity(
                        buffer_element_count,
                    )),
                    input_channel_count,
                    input_channels_to_ignore,
                    output_channel_count,
                }
            }
        }
    }
```

In addition, I'd like to replace the `if .. == ..` pattern by `match` pattern, which can automatically detect the `SampleFormat` cases we forget to handle.

src/backend/buffer_manager.rs
```diff
@@ -58,27 +163,40 @@ impl BufferManager {
                 linear_input_buffer: FloatLinearInputBuffer(Vec::<f32>::with_capacity(
                     buffer_element_count,
                 )),
+                input_channels,
+                input_channels_to_ignore: input_channels_to_ignore as usize,
             }
         }
     }
-    pub fn push_data(&mut self, input_data: *const c_void, read_samples: usize) {
+    pub fn push_data(&mut self, input_data: *mut c_void, frame_count: usize, channel_count: usize) {
```
ChunMinChang:
```suggestion
    pub fn push_data(&mut self, data: *mut c_void, frame_count: usize) {
```

`channel_count` is a fixed value after setup. We can move it to ctor.

src/backend/buffer_manager.rs
```diff
@@ -29,16 +124,24 @@ pub struct BufferManager {
     consumer: RingBufferConsumer,
     producer: RingBufferProducer,
     linear_input_buffer: LinearInputBuffer,
+    // Number of input channels that will be appended to the ring buffer, i.e. the number of
+    // channels that will be presented to the user callback.
+    input_channels: usize,
+    // Number of input channels that needs to be ignored, coming in.
+    input_channels_to_ignore: usize,
```
ChunMinChang, line 130:
```rs
pub struct BufferManager {
    consumer: RingBufferConsumer,
    producer: RingBufferProducer,
    linear_buffer: LinearBuffer,
    // The number of channels in the interleaved data given to push_data
    input_channel_count: usize,
    // The number of channels that needs to be skipped in the beginning of input_channel_count
    input_channels_to_ignore: usize,
    // The number of channels we actually needs, which is also the channel count of the
    // processed data stored in the internal ring buffer.
    output_channel_count: usize,
}
```

To unify the concept of _input_, I'd to have something similar as above

src/backend/mod.rs
```diff
@@ -2558,8 +2591,27 @@ impl<'ctx> CoreStreamData<'ctx> {
...
+                        to_ignore
+                    );
+                    to_ignore
+                } else {
+                    cubeb_log!("Device used as output doesn't have input channels.");
+                    0
+                }
+            } else {
+                0
+            };
```
```suggestion
```
We can use `self.self.input_dev_desc.mChannelsPerFrame - device_channel_count` to get the ignored channel count instead.

src/backend/mod.rs
```diff
@@ -2558,8 +2591,27 @@ impl<'ctx> CoreStreamData<'ctx> {
...
+
             self.input_buffer_manager = Some(BufferManager::new(
                 &self.input_stream_params,
+                input_channels_to_ignore,
                 SAFE_MAX_LATENCY_FRAMES,
             ));
```
```suggestion
           /// When we use the aggregate device, the self.input_dev_desc.mChannelsPerFrame is the
            // total input channel count of all the device added in the aggregate device. However,
            // we only need the audio data captured by the requested input device, so we need to
            // ignore some data captured by the audio input of the requested output device (e.g.,
            // the requested output device is a USB headset with built-in mic), in the beginning of
            // the raw data taken from input callback.
            self.input_buffer_manager = Some(BufferManager::new(
                self.input_stream_params.format(),
                SAFE_MAX_LATENCY_FRAMES as usize,
                self.input_dev_desc.mChannelsPerFrame as usize,
                (self.input_dev_desc.mChannelsPerFrame - device_channel_count) as usize,
                self.input_stream_params.channels() as usize,
            ));
```

src/backend/mod.rs
```diff
@@ -412,10 +420,11 @@ extern "C" fn audiounit_input_callback(
...
             assert_eq!(status, NO_ERR);
-            // Copy input data in linear buffer.
-            let elements =
-                (input_frames * stm.core_stream_data.input_desc.mChannelsPerFrame) as usize;
-            input_buffer_manager.push_data(input_buffer_list.mBuffers[0].mData, elements);
+            input_buffer_manager.push_data(
+                input_buffer_list.mBuffers[0].mData,
+                input_frames as usize,
+                stm.core_stream_data.input_desc.mChannelsPerFrame as usize,
+            );
```
```suggestion
            input_buffer_manager
                .push_data(input_buffer_list.mBuffers[0].mData, input_frames as usize);
```

src/backend/mod.rs
```diff
@@ -508,7 +517,7 @@ fn compute_output_latency(stm: &AudioUnitStream, audio_output_time: u64, now: u6
 
 fn compute_input_latency(stm: &AudioUnitStream, audio_input_time: u64, now: u64) -> u32 {
     const NS2S: u64 = 1_000_000_000;
-    let input_hw_rate = stm.core_stream_data.input_hw_rate as u64;
+    let input_hw_rate = stm.core_stream_data.input_desc.mSampleRate as u64;
```
```suggestion
    let input_hw_rate = stm.core_stream_data.input_dev_desc.mSampleRate as u64;
```

src/backend/mod.rs
```diff
@@ -595,14 +604,14 @@ extern "C" fn audiounit_output_callback(
     let (input_buffer, mut input_frames) = if !stm.core_stream_data.input_unit.is_null() {
         let input_buffer_manager = stm.core_stream_data.input_buffer_manager.as_mut().unwrap();
         assert_ne!(stm.core_stream_data.input_desc.mChannelsPerFrame, 0);
-        let input_channels = stm.core_stream_data.input_desc.mChannelsPerFrame as usize;
+        let input_channels = stm.core_stream_data.input_stream_params.channels() as usize;
```
```suggestion
        let input_channels = input_buffer_manager.channel_count() as usize;
```

We should add a function in `BufferManager` to get this instead:
```rs
impl BufferManager {
    ...
    pub fn channel_count(&self) -> usize {
        self.output_channel_count
    }
    ...
```

src/backend/mod.rs
```diff
@@ -595,14 +604,14 @@ extern "C" fn audiounit_output_callback(
...
         // If the output callback came first and this is a duplex stream, we need to
         // fill in some additional silence in the resampler.
         // Otherwise, if we had more than expected callbacks in a row, or we're
         // currently switching, we add some silence as well to compensate for the
         // fact that we're lacking some input data.
         let input_frames_needed = minimum_resampling_input_frames(
-            stm.core_stream_data.input_hw_rate,
+            stm.core_stream_data.input_desc.mSampleRate,
```
```suggestion
            stm.core_stream_data.input_dev_desc.mSampleRate,
```

src/backend/mod.rs
```diff
@@ -2521,16 +2558,12 @@ impl<'ctx> CoreStreamData<'ctx> {
...
-            // Input AudioUnit must be configured with device's sample rate.
-            // we will resample inside input callback.
-            src_desc.mSampleRate = self.input_hw_rate;
             let r = audio_unit_set_property(
                 self.input_unit,
                 kAudioUnitProperty_StreamFormat,
                 kAudioUnitScope_Output,
                 AU_IN_BUS,
-                &src_desc,
+                &self.input_desc,
```
```suggestion
                &self.input_dev_desc,
```

src/backend/mod.rs
```diff
@@ -2768,7 +2816,7 @@ impl<'ctx> CoreStreamData<'ctx> {
 
         let resampler_input_params = if self.has_input() {
             let mut params = unsafe { *(self.input_stream_params.as_ptr()) };
-            params.rate = self.input_hw_rate as u32;
+            params.rate = self.input_desc.mSampleRate as u32;
```
```suggestion
            params.rate = self.input_dev_desc.mSampleRate as u32;
```

src/backend/mod.rs
```diff
@@ -3513,7 +3576,7 @@ impl<'ctx> StreamOps for AudioUnitStream<'ctx> {
     #[cfg(not(target_os = "ios"))]
     fn input_latency(&mut self) -> Result<u32> {
         let user_rate = self.core_stream_data.input_stream_params.rate();
-        let hw_rate = self.core_stream_data.input_hw_rate as u32;
+        let hw_rate = self.core_stream_data.input_desc.mSampleRate as u32;
```
```suggestion
        let hw_rate = self.core_stream_data.input_dev_desc.mSampleRate as u32;
```

ChunMinChang (changes requested):
If we have no way to set the channel order for the aggregate device, using a remixer in BufferManager seems a feasible way to go.

The idea looks good. The proposed changes inspires me some ideas to make the code cleaner.

I have no way to add the _suggestion_ block for the unchanged code so I put all my suggestion here: https://github.com/ChunMinChang/cubeb-coreaudio-rs/commit/6408490256642f058e560ce32c2b14c70a39b55b

src/backend/tests/device_change.rs
ChunMinChang: I am not sure if we still need this. The test passed in https://github.com/ChunMinChang/cubeb-coreaudio-rs/commit/6408490256642f058e560ce32c2b14c70a39b55b

ChunMinChang (changes requested):

src/backend/mod.rs
padenot: For the output side, it's well-defined, we have a mixer that will down-mix properly because there are mapping information (it's useful to play, say, 5.1 audio on stereo). For input, definitely yes, your version is better.

src/backend/tests/device_change.rs
padenot: I'll back out and see.

src/backend/mod.rs
padenot: They are different, I've added a comment.

src/backend/mod.rs
padenot: What's the difference with what we have? We need both the channel count of the aggregate, and the input device, that doesn't really help.

src/backend/mod.rs
padenot: Yes, this is the point. I've renamed some variables to make this clearer.

padenot:
Well this branch has my fix.

padenot:
You know you can push to other people's PR directly, yes? It's a lot easier this way.

I've actually reverted my patch that fixes the `unwrap`, here is a failure: https://github.com/mozilla/cubeb-coreaudio-rs/runs/5601532591?check_suite_focus=true#step:6:526

ChunMinChang (approved):
It seems we finally fix the mystery crash in `test_destroy_duplex_stream_after_unplugging_a_nondefault_input_device` by no adding one aggregate device into another aggregate device? Nice!

PR #158, opened 2022-03-18
Close the stream before entering error state

ChunMinChang:
The audio stream will enter an error state once its device is unplugged. But before firing the error callback, we should stop its AudioUnits to stop getting the unnecessary data callbacks.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -752,8 +752,9 @@ extern "C" fn audiounit_property_listener_callback(
     // Handle the events
     if input_device_dead {
         cubeb_log!("The user-selected input device is dead, enter error state");
-        stm.notify_state_changed(State::Error);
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -3279,6 +3280,25 @@ impl<'ctx> AudioUnitStream<'ctx> {
         });
     }
 
+    fn close_on_error(&mut self) {
+        let queue = self.queue.clone();
+        let mutexed_stm = Arc::new(Mutex::new(self));
+        let also_mutexed_stm = Arc::clone(&mutexed_stm);
+
+        // Use a new thread, through the queue, to avoid deadlock when calling
```
padenot: "different", not "new", right? 
```suggestion
        // Use a different thread, through the queue, to avoid deadlock when calling
```

padenot (approved):

PR #161, opened 2022-03-22
Restore the ability to open a stream with more channels than the output device has channels

padenot:
The mixer handles it. This allows playing e.g. a 5.1 file correctly on 2.0 hardware.

This reverts part of https://github.com/mozilla/cubeb-coreaudio-rs/pull/153/commits/8d7d78637d839d8b9e0be04762ea964999a6091f#diff-b1bbc13b52b920f2cfb842559dcccfa6a836168e2061e22d8a897c82af70c3a7R2645, so that we don't return an error, we simply set up the mixer correctly.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -2614,12 +2614,6 @@ impl<'ctx> CoreStreamData<'ctx> {
                 out_dev_info
             );
 
-            let device_channel_count =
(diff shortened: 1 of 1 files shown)


padenot:
Merging this now so I can land the whole thing in gecko.

PR #162, opened 2022-03-25
Use the user-provided rate to open the output device, and not the hardware rate

padenot:
Fix for [BMO#1761392](https://bugzilla.mozilla.org/show_bug.cgi?id=1761392).


padenot:
Merging now so I can make the next Nightly for sure.

PR #163, opened 2022-03-25
When the default output or input are selected, and the device disappears, the new defaults should be used

padenot:


src/backend/mod.rs
```diff
@@ -3286,6 +3286,40 @@ impl<'ctx> AudioUnitStream<'ctx> {
...
 
+        // Use the new default device if this stream was set to follow the output device.
+        if self.core_stream_data.has_output()
+            && self
+                .core_stream_data
+                .output_device
+                .flags
+                .contains(device_flags::DEV_SELECTED_DEFAULT)
+        {
+            self.core_stream_data.output_device = create_device_info(kAudioObjectUnknown, DeviceType::OUTPUT).map_err(|e| {
```
ChunMinChang, line 3297: It would be better to change the returned value of `audiounit_get_default_device_id` to `Result<AudioObjectID, OSStatus>` so we can just set the `self.core_stream_data.output_device.id` here directly.

src/backend/mod.rs
```diff
@@ -3286,6 +3286,40 @@ impl<'ctx> AudioUnitStream<'ctx> {
...
+
+        // Likewise, for the input seid
+        if self.core_stream_data.has_input()
+            && self
+                .core_stream_data
+                .input_device
+                .flags
+                .contains(device_flags::DEV_SELECTED_DEFAULT)
+        {
+            self.core_stream_data.input_device = create_device_info(kAudioObjectUnknown, DeviceType::INPUT).map_err(|e| {
```
ChunMinChang, line 3314: Same as output.

ChunMinChang (approved):
LGTM, thanks!

PR #164, opened 2022-03-25
Don't multiply by zero when getting the input latency with different hw vs. user rates

padenot:


src/backend/mod.rs
```diff
@@ -3526,7 +3526,7 @@ impl<'ctx> StreamOps for AudioUnitStream<'ctx> {
             if hw_rate == user_rate {
                 Ok(frames)
             } else {
-                Ok(frames * (user_rate / hw_rate))
+                Ok((frames * user_rate) / hw_rate)
```
ChunMinChang, line 3529: It would be great if we have a test for this.

ChunMinChang (approved):
LGTM, thanks!

PR #165, opened 2022-03-25
Some cleanup

ChunMinChang:
- Change returned values of `audiounit_get_default_device_id` and its friends
- Simplify `create_device_info`

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -189,31 +189,27 @@ fn clamp_latency(latency_frames: u32) -> u32 {
     )
 }
 
-fn create_device_info(id: AudioDeviceID, devtype: DeviceType) -> Result<device_info> {
-    assert_ne!(id, kAudioObjectSystemObject);
+fn create_device_info(
+    devid: AudioDeviceID,
+    devtype: DeviceType,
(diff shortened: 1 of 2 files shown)


src/backend/aggregate_device.rs
```diff
@@ -475,7 +475,7 @@ impl AggregateDevice {
         };
 
         // Master become the 1st output sub device
-        let output_device_id = audiounit_get_default_device_id(DeviceType::OUTPUT);
+        let output_device_id = audiounit_get_default_device_id(DeviceType::OUTPUT).unwrap();
```
padenot: This crashes when there are no output devices, right? Is this what we want?

padenot: Probably fixed in the other PR anyways.

padenot (approved):

src/backend/aggregate_device.rs
ChunMinChang: This is fixed in #166.

We won't create aggregate device if there is no output device. The aggregate device is always created when we have both input and output device. However, this is still wrong, since the aggregate device's the output device can be different than the default output device.

PR #166, opened 2022-03-25
Correct the master device of the aggregate device

ChunMinChang:
We should set the master device of the aggregate device to the output
device of the aggregate device, instead of the default output device.


padenot (approved):
Very cool, thanks.

PR #167, opened 2022-03-28
 Return error for default device if it's `kAudioObjectUnknown`

ChunMinChang:
Return error for default device if it's `kAudioObjectUnknown`

The default output device on Circle CI server is `kAudioObjectUnknown`.
We should return error if we get this kind of device since all its
operations will be invalid.


kinetiknz (approved):
LGTM

PR #170, opened 2022-03-31
Watching the alive status for all the running devices

### LINKED ISSUE #159: Fire error callback when non-default output device is unplugged
Now we disable the automatic input device switching when the input device is unplugged. We will fire an error callback and enter an error state if this happens. However, we never monitor the alive status for output device. If the output device doesn't follow the default, then the stream will probably stop working without firing an error callback. Now the cubeb users rely on device-collection changes to know the device is gone.

To align with the policy we have for microphone switching, we should do the same for the output device.

ChunMinChang:
We should monitor the alive status of both input and output devices in use. We used to watch input device only. With the code in this PR, we are able to watch the alive status of all the running devices when user doesn't follow the system default devices. If any user-selected device is unplugged before it's destroyed, we should fire an error callback.

### DIFF
--- a/run_device_tests.sh
+++ b/run_device_tests.sh
@@ -24,10 +24,8 @@ cargo test test_suspend_input_stream_by_unplugging_a_nondefault_input_device -- 
 cargo test test_destroy_input_stream_after_unplugging_a_default_input_device -- --ignored --nocapture
 cargo test test_reinit_input_stream_by_unplugging_a_default_input_device -- --ignored --nocapture
 
-# FIXME: We don't monitor the alive-status for output device currently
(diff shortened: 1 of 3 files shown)


padenot:
Does this solve a problem in practice?

ChunMinChang:
> Does this solve a problem in practice?

I believe this relates to [setSinkId](https://w3c.github.io/mediacapture-output/#algorithms-sink-unavailable). Now if the user-selected/non-default output device is unplugged while it's playing, the underlying stream will stop playing behind the scenes without notifying gecko to tear down the related variables or settings.

In addition, cubeb already does this job in WASAPI and PulseAudio backends so this makes cubeb behaves the same way on CoreAudio backend.

padenot:
Sure, ok, I really wasn't sure what this was about, let me know when it's ready!

PR #171, opened 2022-04-06
Allow passing an unknown object to get_device_transport_type

padenot:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1762563#c1

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -434,6 +434,12 @@ extern "C" fn audiounit_input_callback(
             ptr::null_mut(),
             0,
         );
+        if outframes < 0 {
+            stm.stopped.store(true, Ordering::SeqCst);
+            stm.core_stream_data.stop_audiounits();
+            stm.notify_state_changed(State::Error);
+            return handle;
+        }
         if outframes < total_input_frames {
             stm.draining.store(true, Ordering::SeqCst);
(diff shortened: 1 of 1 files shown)


ChunMinChang (commented):
It seems this only happens during reinit and some machines return unknown as their default device (when they actually don't have any device?). I think we should revert what I did in #167 and do something like the following in `reinit`. What do you think?

```rs
        if self.core_stream_data.has_output()
            && self
                .core_stream_data
                .output_device
                .flags
                .contains(device_flags::DEV_SELECTED_DEFAULT)
        {
            self.core_stream_data.output_device = match create_device_info(kAudioObjectUnknown, DeviceType::OUTPUT) {
                None => {
                    cubeb_log!("Fail to create device info for output");
                    return Err(Error::error());
                }
                Some(d) => d,
                };
        }

```

padenot:
Yeah we thought we'd already handle this case: https://github.com/mozilla/cubeb-coreaudio-rs/blob/39f7e696c11ff45503b50a49a36cd9f39900c27b/src/backend/mod.rs#L3309-L3324. I'm not sure I can verify, doing the equivalent of `sudo kextunload /System/Library/Extensions/AppleHDA.kext` doesn't work on my new macbook.

PR #173, opened 2022-04-12
Fix device reinit when no input/output device are present

padenot:
Per https://github.com/mozilla/cubeb-coreaudio-rs/pull/171#pullrequestreview-933864519.

@ChunMinChang: First commit really is just the revert / merge. Second one implements your suggestion:

### DIFF
--- a/src/backend/device_property.rs
+++ b/src/backend/device_property.rs
@@ -4,6 +4,8 @@ pub fn get_device_uid(
     id: AudioDeviceID,
     devtype: DeviceType,
 ) -> std::result::Result<StringRef, OSStatus> {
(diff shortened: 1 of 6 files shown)


padenot:
What do you want to do about this failure now? Can we just stop asserting everything?

ChunMinChang:
I believe the reason you need to remove all the assert is that you remove the `get_default_device` as well. Using `match create_device_info(kAudioObjectUnknown, DeviceType::XXX)` for reinit should achieve the goal I guess

ChunMinChang:
I am going to merge this since it should fix the problem now.

PR #174, opened 2022-04-12
Don't assert about input_frames / total_input_frames when input-only.

padenot:
This is possible if the first branch of:

> if status == kAudioUnitErr_CannotDoInCurrentContext

is taken and there's been data before. In this case, it's possible to
have input frames, but impossible to have them, probably. I can
reproduce 100% of the time by installing BlackHole-16ch and running `./run_tests.sh` on the M1 Max.

### DIFF
--- a/src/backend/buffer_manager.rs
+++ b/src/backend/buffer_manager.rs
@@ -179,7 +179,7 @@ impl BufferManager {
             }
         }
     }
-    pub fn channel_count(&self) -> usize {
+    fn channel_count(&self) -> usize {
         self.output_channel_count
     }
     pub fn push_data(&mut self, data: *mut c_void, frame_count: usize) {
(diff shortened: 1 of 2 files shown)


ChunMinChang (changes requested):
> This is possible if the first branch of:
> 
> > if status == kAudioUnitErr_CannotDoInCurrentContext
> 
> is taken and there's been data before. In this case, it's possible to have input frames, but impossible to have them, probably. I can reproduce 100% of the time by installing BlackHole-16ch and running `./run_tests.sh` on the M1 Max.

I am not sure I understand why `assert!(input_frames as i64 <= total_input_frames)` will be hit without hitting `assert!(!stm.core_stream_data.output_unit.is_null())` first:
https://github.com/mozilla/cubeb-coreaudio-rs/blob/39f7e696c11ff45503b50a49a36cd9f39900c27b/src/backend/mod.rs#L386-L387

If code path of `status == kAudioUnitErr_CannotDoInCurrentContext` can happen when stream is input-only, we should remove 
1. `assert!(!stm.core_stream_data.output_unit.is_null())`
2. append the silence to `input_buffer_manager` if it's a input-only stream. (via `get_linear_data`?). Otherwise we won't have enough frame for the calback.

```rs
        let handle = if status == kAudioUnitErr_CannotDoInCurrentContext {
            if !stm.core_stream_data.output_unit.is_null() {
                // kAudioUnitErr_CannotDoInCurrentContext is returned when using a BT
                // headset and the profile is changed from A2DP to HFP/HSP. The previous
                // output device is no longer valid and must be reset.
                // For now state that no error occurred and feed silence, stream will be
                // resumed once reinit has completed.
                cubeb_logv!(
                    "({:p}) input: reinit pending, output will pull silence instead",
                    stm.core_stream_data.stm_ptr
                );
                ErrorHandle::Reinit
            } else {
                // Push input_frames silence data to input_buffer_manager...
            }
        } else {
            assert_eq!(status, NO_ERR);
            input_buffer_manager
                .push_data(input_buffer_list.mBuffers[0].mData, input_frames as usize);
            ErrorHandle::Return(status)
        };
```

ChunMinChang:
After testing `BlackHole-16ch` on my MBP M1 Max, I find the problem is in input frame calculation instead of that assertion itself. I've pushed some code in this PR so you can test it on your end to see if the problem solves or not.

padenot:
> I am not sure I understand why `assert!(input_frames as i64 <= total_input_frames)` will be hit without hitting `assert!(!stm.core_stream_data.output_unit.is_null())` first:

I think it's related with Siri or concurrent device reconfiguration maybe? You have the input callback firing with a number of frames, but when you attempt to render, it says it's now unavailable. We've hit this before (hence this code path), and I'm pretty sure it was at least when doing the device had an input stream going, and siri was triggered.

Anyways, yeah the `buffer_manager` code was wrong, thanks for the fix, it works locally as well for me.

padenot:
Cleaned up the branch and merging.

PR #177, opened 2022-07-22
Allow failure of current_device

ChunMinChang:
`AudioUnitStream::current_device` can fail when the default device has
no `kAudioDevicePropertyDataSource` value.


ChunMinChang:
We got a compiler error with Rust Nightly. See https://github.com/rust-lang/rust/issues/99563

kinetiknz (approved):

PR #178, opened 2022-09-13
Don't call `set_notification_runloop`

ChunMinChang:
To fix the issue mentioned in [bug 1781964 comment 12](https://bugzilla.mozilla.org/show_bug.cgi?id=1781964#c12) and, we should stop calling `set_notification_runloop` on Mac OS 10.7

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -164,24 +164,6 @@ impl From<CAChannelLabel> for mixer::Channel {
     }
 }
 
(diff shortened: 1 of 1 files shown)


ChunMinChang:
`os_version` relies on `plist` crate. I am considering using other method to get the os version without the dependencies so I won't need to audit `plist` in gecko.

padenot:
Yeah we don't care about super old macOS here, 10.7 is ancient.

PR #180, opened 2022-09-23
Don't log from other threads than the output thread, in duplex mode

padenot:
This uses an spsc ring buffer to proxy the audio input callback parameters to the output thread, for logging purposes.

Without this, multiple threads can attempt writing in the spsc ring buffer that's used for async logging, which make an assertion blow up.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -32,6 +32,7 @@ use self::mixer::*;
 use self::resampler::*;
 use self::utils::*;
 use atomic;
+use backend::ringbuf::RingBuffer;
 use cubeb_backend::{
     ffi, Context, ContextOps, DeviceCollectionRef, DeviceId, DeviceRef, DeviceType, Error, Ops,
     Result, SampleFormat, State, Stream, StreamOps, StreamParams, StreamParamsRef, StreamPrefs,
@@ -39,6 +40,7 @@ use cubeb_backend::{
 use mach::mach_time::{mach_absolute_time, mach_timebase_info};
 use std::cmp;
 use std::ffi::{CStr, CString};
+use std::fmt;
 use std::mem;
 use std::os::raw::c_void;
 use std::ptr;
@@ -390,10 +392,6 @@ extern "C" fn audiounit_input_callback(
             // output device is no longer valid and must be reset.
             // For now state that no error occurred and feed silence, stream will be
             // resumed once reinit has completed.
-            cubeb_alog!(
-                "({:p}) input: reinit pending, output will pull silence instead",
-                stm.core_stream_data.stm_ptr
-            );
             ErrorHandle::Reinit
         } else {
             assert_eq!(status, NO_ERR);
@@ -401,6 +399,24 @@ extern "C" fn audiounit_input_callback(
                 .push_data(input_buffer_list.mBuffers[0].mData, input_frames as usize);
             ErrorHandle::Return(status)
         };
+
+        // Full Duplex. We'll call data_callback in the AudioUnit output callback. Record this
+        // callback for logging.
+        if !stm.core_stream_data.output_unit.is_null() {
+            let input_callback_data = InputCallbackData {
+                bytes: input_buffer_list.mBuffers[0].mDataByteSize,
+                rendered_frames: input_frames,
+                total_available: input_buffer_manager.available_frames(),
+                channels: input_buffer_list.mBuffers[0].mNumberChannels,
+                num_buf: input_buffer_list.mNumberBuffers,
+            };
+            stm.core_stream_data
+                .input_logging
+                .as_mut()
+                .unwrap()
+                .push(input_callback_data);
+            return handle;
+        }
 
         cubeb_alogv!(
             "({:p}) input: buffers {}, size {}, channels {}, rendered frames {}, total frames {}.",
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -2234,6 +2267,33 @@ struct CoreStreamData<'ctx> {
     input_alive_listener: Option<device_property_listener>,
     input_source_listener: Option<device_property_listener>,
     output_source_listener: Option<device_property_listener>,
+    input_logging_prod: Option<ringbuf::Producer<InputCallbackData>>,
+    input_logging_cons: Option<ringbuf::Consumer<InputCallbackData>>,
+}
+
+impl<'ctx> fmt::Debug for CoreStreamData<'ctx> {
```
padenot: I don't know how to make this work with `#[derive(Debug)]` otherwise: `ringbuf::Consumer<...>` doesn't have `Debug`.

src/backend/mod.rs
```diff
@@ -635,6 +651,15 @@ extern "C" fn audiounit_output_callback(
         (ptr::null_mut::<c_void>(), 0)
     };
 
+    cubeb_alogv!(
```
padenot, line 654: We always log input data earlier, then output data.

src/backend/mod.rs
```diff
@@ -581,6 +582,21 @@ extern "C" fn audiounit_output_callback(
 
     // Also get the input buffer if the stream is duplex
     let (input_buffer, mut input_frames) = if !stm.core_stream_data.input_unit.is_null() {
+        let rb_cons = stm.core_stream_data.input_logging_cons.as_mut().unwrap();
+        if rb_cons.is_empty() {
+            cubeb_alogv!("no audio input data in output callback");
```
padenot, line 587: This allows getting something in the log when there's something going on with the input side, for example the log that I removed in the input callback:
```rust
cubeb_alog!(
  "({:p}) input: reinit pending, output will pull silence instead",
  stm.core_stream_data.stm_ptr
);
```

src/backend/mod.rs
```diff
@@ -402,6 +400,24 @@ extern "C" fn audiounit_input_callback(
...
+            };
+            stm.core_stream_data
+                .input_logging_prod
+                .as_mut()
+                .unwrap()
+                .push(input_callback_data);
+            return handle;
+        }
+
         cubeb_alogv!(
```
padenot, line 421: Log normally if not duplex: there's going to be a single thread that log.

src/backend/mod.rs
```diff
@@ -2234,6 +2267,33 @@ struct CoreStreamData<'ctx> {
     input_alive_listener: Option<device_property_listener>,
     input_source_listener: Option<device_property_listener>,
     output_source_listener: Option<device_property_listener>,
+    input_logging_prod: Option<ringbuf::Producer<InputCallbackData>>,
+    input_logging_cons: Option<ringbuf::Consumer<InputCallbackData>>,
```
ChunMinChang: It's better to have one struct managing this pair so it's clear to see `input_logging_prod` and `input_logging_cons` should be created or destroyed at the same time, and also decrease the mind complexity to know when prod or cons should be used.

src/backend/mod.rs
ChunMinChang: I would prefer something like:

```rust
struct InputCallbackLogger {
    prod: ringbuf::Producer<InputCallbackData>,
    cons: ringbuf::Consumer<InputCallbackData>,
}

impl InputCallbackLogger {
    fn new() -> Self {
        let ring = RingBuffer::<InputCallbackData>::new(16);
        let (prod, cons) = ring.split();
        Self { prod, cons }
    }

    fn push(&mut self, data: InputCallbackData) {
        self.prod.push(data);
    }

    fn pop(&mut self) -> Option<InputCallbackData> {
        self.cons.pop()
    }

    fn is_empty(&self) -> bool {
        self.cons.is_empty()
    }
}

impl fmt::Debug for InputCallbackLogger {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        // something useful for debugging?
        write!(
            f,
            "InputCallbackLogger  {{ prod: {}, cons: {} }}",
            self.prod.len(),
            self.cons.len()
        )
    }
}
```

ChunMinChang (commented):
This seems to me it's better to replace the spsc-queue in the async log by a mpsc-queue like this: https://searchfox.org/mozilla-central/rev/d45dd05bf412e7468b3770a52519e9d546d6325c/dom/media/MPSCQueue.h#10. Otherwise we need to do this for all the backends. That's what gecko does for async log: https://searchfox.org/mozilla-central/rev/d45dd05bf412e7468b3770a52519e9d546d6325c/dom/media/AsyncLogger.h

What do you think?

padenot:
An MPSC that doesn't allocate is hard to write, and we'd need it in C++ (the gecko one allocates, and we do a number of tricks to make it cheap). When we have more backends in rust, we can reconsider what we do about logging, probably by using a good MPSC crate.

ChunMinChang (approved):
Look fine to me.

Using multi-producer techniques will be way better than this workaround. 

Facebook's *folly* has some example code implementing multi-producer without allocating memory when pushing data: https://github.com/facebook/folly/blob/fe37daacf861700585d303b277138cb18a92686c/folly/MPMCQueue.h

Here is another example: https://github.com/cameron314/concurrentqueue

Please open a ticket to track this issue since we will have this problem again if we need to do the same thing on other backends or log some messages asynchronously from notification threads.

PR #182, opened 2022-10-18
Use serial queue to stop AudioUnits when handling errors in a render callback.

kinetiknz:
Avoids a possible deadlock inside CoreAudio if another thread calls AudioOutputUnitStop concurrently.  The thread running the render callback holds the HALB_Mutex and blocks acquiring the CAMutex inside AudioOutputUnitStop. The racing thread calling AudioOutputUnitStop holds the CAMutex and blocks acquiring the HALB_Mutex.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -439,8 +439,16 @@ extern "C" fn audiounit_input_callback(
         );
         if outframes < 0 {
             stm.stopped.store(true, Ordering::SeqCst);
-            stm.core_stream_data.stop_audiounits();
             stm.notify_state_changed(State::Error);
+            let queue = stm.queue.clone();
+            let mutexed_stm = Arc::new(Mutex::new(stm));
+            let also_mutexed_stm = Arc::clone(&mutexed_stm);
+            // Use a new thread, through the queue, to avoid deadlock when calling
+            // AudioOutputUnitStop method from inside render callback
+            queue.run_async(move || {
+                let stm_guard = also_mutexed_stm.lock().unwrap();
+                stm_guard.core_stream_data.stop_audiounits();
+            });
             return handle;
         }
         if outframes < total_input_frames {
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -669,9 +677,17 @@ extern "C" fn audiounit_output_callback(
 
     if outframes < 0 || outframes > i64::from(output_frames) {
         stm.stopped.store(true, Ordering::SeqCst);
-        stm.core_stream_data.stop_audiounits();
-        audiounit_make_silent(&mut buffers[0]);
         stm.notify_state_changed(State::Error);
```
ChunMinChang, line 680: Should the state notification be fired after silencing the output buffer?

ChunMinChang (commented):
I'd like to figure out what happen exactly first. So there are two threads. One holds *HALB_Mutex* and the other holds *CAMutex*. I'd like to know what thread, or which call, holds or tries to acquire *HALB_Mutex* and  *CAMutex*. Does the thread running input or output callback hold the *HALB_Mutex*? Does calling `stop_audiounits` lead to *CAMutex* acquisition? Is it possible to write a test to simulate it?

kinetiknz:
Gecko thread stack:
```
 0  libsystem_kernel.dylib!__psynch_mutexwait
 1  libsystem_pthread.dylib!_pthread_mutex_firstfit_lock_slow
 2  CoreAudio!HALB_Mutex::Lock
 3  CoreAudio!HALC_ProxyIOContext::StopIOProc
 4  CoreAudio!HAL_HardwarePlugIn_DeviceStop
 5  CoreAudio!HALDevice::StopIOProc
 6  CoreAudio!AudioDeviceStop
 7  CoreAudio
 8  CoreAudio
 9  CoreAudio
10  XUL!coreaudio_sys_utils::audio_unit::audio_output_unit_stop
11  XUL!cubeb_coreaudio::backend::stop_audiounit
12  XUL!cubeb_coreaudio::backend::CoreStreamData::stop_audiounits
13  XUL!<cubeb_coreaudio::backend::AudioUnitStream as cubeb_backend::traits::StreamOps>::stop::{{closure}}
```

CoreAudio render thread (com.apple.audio.IOThread.client):
```
 0  libsystem_kernel.dylib!__psynch_mutexwait
 1  libsystem_pthread.dylib!_pthread_mutex_firstfit_lock_slow
 2  libAudioToolboxUtility.dylib!CAMutex::Lock
 3  CoreAudio
 4  XUL!coreaudio_sys_utils::audio_unit::audio_output_unit_stop
 5  XUL!cubeb_coreaudio::backend::stop_audiounit
 6  XUL!cubeb_coreaudio::backend::CoreStreamData::stop_audiounits
 7  XUL!cubeb_coreaudio::backend::audiounit_output_callback
 8  CoreAudio
 9  CoreAudio
10  CoreAudio
11  CoreAudio!invocation function for block in HALC_ProxyIOContext::HALC_ProxyIOContext
12  CoreAudio!HALB_IOThread::Entry
```

CoreAudio's render thread holds HALB_Mutex during our audiounit_output_callback, then blocks on CAMutex.  The Gecko thread holds CAMutex (inside AudioOutputUnitStop), then blocks on HALB_Mutex.

It'd be quite difficult to write a useful test for this as the the incorrect code would still pass whenever the stop from inside the render callback completed the CAMutex section of its work before the stop from another thread acquires CAMutex.

src/backend/mod.rs
kinetiknz: The buffer wouldn't be used until we return from the render callback, so I can't see a reason for the ordering to matter.  Can you expand on your concern here?

ChunMinChang: It seems the order shouldn't make a difference

ChunMinChang (approved):

PR #183, opened 2022-10-18
Remove also_mutexed_stm pattern when making serial queue calls.

kinetiknz:
The lifetime of the stream is guaranteed to outlive any items running on the
serial queue.  This is ensured by stream destroy flushing the serial queue
before freeing stream resources.

The `Arc<Mutex<&mut AudioUnitStream>>` seems to be unnecessary.

This PR is just the second commit, but is based on https://github.com/mozilla/cubeb-coreaudio-rs/pull/182


ChunMinChang (approved):
LGTM, thanks!

This pattern was a workaround to keep using the original variable but allow moving the cloned reference to a closure. But this pattern is unnecessary and redundant now.

PR #184, opened 2023-07-25
Bump bitflags to version 2

padenot:
Like all other cubeb crates


ChunMinChang (approved):
LGTM, and I've fixed some clippy issues so it's good to land

PR #185, opened 2023-10-11
Use a VoiceProcessingIO AudioUnit with bypass enabled instead of a regular AudioUnit on top of a drift-corrected aggregate device when in VOICE mode

Pehrsons:
The drift corrected aggregate device is resulting in distortions due to drift, and we do not really understand why.
The VoiceProcessingIO AudioUnit handles drift properly.

This is the workaround fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1670633.

Note that the VoiceProcessingIO AudioUnit enables system wide ducking on other output streams while running.
We set it to bypass mode for now since there is not yet an API to toggle processing features dynamically, as would be needed by Firefox through MediaStreamTrack.applyConstraints.

### DIFF
--- a/run_device_tests.sh
+++ b/run_device_tests.sh
@@ -31,7 +31,7 @@ cargo test test_destroy_output_stream_after_unplugging_a_default_output_device -
 cargo test test_reinit_output_stream_by_unplugging_a_default_output_device -- --ignored --nocapture
 
 cargo test test_destroy_duplex_stream_after_unplugging_a_nondefault_input_device -- --ignored --nocapture
-cargo test test_suspend_duplex_stream_by_unplugging_a_nondefault_input_device
+cargo test test_suspend_duplex_stream_by_unplugging_a_nondefault_input_device -- --ignored --nocapture
 
 cargo test test_destroy_duplex_stream_after_unplugging_a_nondefault_output_device -- --ignored --nocapture
 cargo test test_suspend_duplex_stream_by_unplugging_a_nondefault_output_device -- --ignored --nocapture
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -2851,7 +2851,9 @@ impl<'ctx> CoreStreamData<'ctx> {
                 self.stm_ptr,
                 input_hw_desc
             );
-            assert!(input_hw_desc.mChannelsPerFrame >= device_channel_count);
+            // In some cases with VPIO the stream format's mChannelsPerFrame is higher than
+            // expected. Use get_channel_count as source of truth.
+            input_hw_desc.mChannelsPerFrame = device_channel_count;
             // Notice: when we are using aggregate device, the input_hw_desc.mChannelsPerFrame is
             // the total of all the input channel count of the devices added in the aggregate device.
             // Due to our aggregate device settings, the data captured by the output device's input
@@ -3028,6 +3030,10 @@ impl<'ctx> CoreStreamData<'ctx> {
                 self.stm_ptr,
                 output_hw_desc
             );
+
+            // In some cases with VPIO the stream format's mChannelsPerFrame is higher than
+            // expected. Use get_channel_count as source of truth.
+            output_hw_desc.mChannelsPerFrame = device_channel_count;
 
             // This has been observed in the wild.
             if output_hw_desc.mChannelsPerFrame == 0 {
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -2134,6 +2134,16 @@ impl ContextOps for AudioUnitContext {
             return Err(Error::invalid_parameter());
         }
 
+        if input_stream_params.is_none() && output_stream_params.is_none() {
+            cubeb_log!("Cannot init a stream without any stream params");
+            return Err(Error::invalid_parameter());
+        }
+
+        if data_callback.is_none() {
+            cubeb_log!("Cannot init a stream without a data callback");
+            return Err(Error::invalid_parameter());
+        }
+
         // Latency cannot change if another stream is operating in parallel. In this case
         // latency is set to the other stream value.
         let global_latency_frames = self
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -1711,19 +1711,6 @@ fn create_cubeb_device_info(
     Ok(dev_info)
 }
 
-fn is_aggregate_device(device_info: &ffi::cubeb_device_info) -> bool {
-    assert!(!device_info.friendly_name.is_null());
-    let private_name =
-        CString::new(PRIVATE_AGGREGATE_DEVICE_NAME).expect("Fail to create a private name");
-    unsafe {
-        libc::strncmp(
-            device_info.friendly_name,
-            private_name.as_ptr(),
-            private_name.as_bytes().len(),
-        ) == 0
-    }
-}
-
 fn destroy_cubeb_device_info(device: &mut ffi::cubeb_device_info) {
     // This should be mapped to the memory allocation in `create_cubeb_device_info`.
     // The `device_id`, `group_id`, `vendor_name` can be null pointer if the queries
@@ -2189,9 +2176,7 @@ impl ContextOps for AudioUnitContext {
             let devices = audiounit_get_devices_of_type(*dev_type);
             for device in devices {
                 if let Ok(info) = create_cubeb_device_info(device, *dev_type) {
-                    if !is_aggregate_device(&info) {
-                        device_infos.push(info);
-                    }
+                    device_infos.push(info);
                 }
             }
         }
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -325,6 +325,7 @@ extern "C" fn audiounit_input_callback(
 
     assert!(!user_ptr.is_null());
     let stm = unsafe { &mut *(user_ptr as *mut AudioUnitStream) };
+    let using_voice_processing_unit = stm.core_stream_data.using_voice_processing_unit();
 
     if unsafe { *flags | kAudioTimeStampHostTimeValid } != 0 {
         let now = unsafe { mach_absolute_time() };
@@ -467,9 +468,7 @@ extern "C" fn audiounit_input_callback(
     // If the input (input-only stream) or the output is drained (duplex stream),
     // cancel this callback. Note that for voice processing cases (a single unit),
     // the output callback handles stopping the unit and notifying of state.
-    if stm.core_stream_data.input_unit != stm.core_stream_data.output_unit
-        && stm.draining.load(Ordering::SeqCst)
-    {
+    if !using_voice_processing_unit && stm.draining.load(Ordering::SeqCst) {
         let r = stop_audiounit(stm.core_stream_data.input_unit);
         assert!(r.is_ok());
         // Only fire state-changed callback for input-only stream.
@@ -834,14 +833,14 @@ fn get_default_device_id(devtype: DeviceType) -> std::result::Result<AudioObject
     }
 }
 
-fn audiounit_convert_channel_layout(layout: &AudioChannelLayout) -> Vec<mixer::Channel> {
+fn audiounit_convert_channel_layout(layout: &AudioChannelLayout) -> Result<Vec<mixer::Channel>> {
     if layout.mChannelLayoutTag != kAudioChannelLayoutTag_UseChannelDescriptions {
         // kAudioChannelLayoutTag_UseChannelBitmap
         // kAudioChannelLayoutTag_Mono
         // kAudioChannelLayoutTag_Stereo
         // ....
         cubeb_log!("Only handling UseChannelDescriptions for now.\n");
-        return Vec::new();
+        return Err(Error::error());
     }
 
     let channel_descriptions = unsafe {
@@ -857,10 +856,10 @@ fn audiounit_convert_channel_layout(layout: &AudioChannelLayout) -> Vec<mixer::C
         channels.push(label.into());
     }
 
-    channels
+    Ok(channels)
 }
 
-fn audiounit_get_preferred_channel_layout(output_unit: AudioUnit) -> Vec<mixer::Channel> {
+fn audiounit_get_preferred_channel_layout(output_unit: AudioUnit) -> Result<Vec<mixer::Channel>> {
     let mut rv = NO_ERR;
     let mut size: usize = 0;
     rv = audio_unit_get_property_info(
@@ -876,7 +875,7 @@ fn audiounit_get_preferred_channel_layout(output_unit: AudioUnit) -> Vec<mixer::
             "AudioUnitGetPropertyInfo/kAudioDevicePropertyPreferredChannelLayout rv={}",
             rv
         );
-        return Vec::new();
+        return Err(Error::error());
     }
     debug_assert!(size > 0);
 
@@ -894,7 +893,7 @@ fn audiounit_get_preferred_channel_layout(output_unit: AudioUnit) -> Vec<mixer::
             "AudioUnitGetProperty/kAudioDevicePropertyPreferredChannelLayout rv={}",
             rv
         );
-        return Vec::new();
+        return Err(Error::error());
     }
 
     audiounit_convert_channel_layout(layout.as_ref())
@@ -902,7 +901,7 @@ fn audiounit_get_preferred_channel_layout(output_unit: AudioUnit) -> Vec<mixer::
 
 // This is for output AudioUnit only. Calling this by input-only AudioUnit is prone
 // to crash intermittently.
-fn audiounit_get_current_channel_layout(output_unit: AudioUnit) -> Vec<mixer::Channel> {
+fn audiounit_get_current_channel_layout(output_unit: AudioUnit) -> Result<Vec<mixer::Channel>> {
     let mut rv = NO_ERR;
     let mut size: usize = 0;
     rv = audio_unit_get_property_info(
@@ -918,8 +917,7 @@ fn audiounit_get_current_channel_layout(output_unit: AudioUnit) -> Vec<mixer::Ch
             "AudioUnitGetPropertyInfo/kAudioUnitProperty_AudioChannelLayout rv={}",
             rv
         );
-        // This property isn't known before macOS 10.12, attempt another method.
-        return audiounit_get_preferred_channel_layout(output_unit);
+        return Err(Error::error());
     }
     debug_assert!(size > 0);
 
@@ -937,10 +935,23 @@ fn audiounit_get_current_channel_layout(output_unit: AudioUnit) -> Vec<mixer::Ch
             "AudioUnitGetProperty/kAudioUnitProperty_AudioChannelLayout rv={}",
             rv
         );
-        return Vec::new();
+        return Err(Error::error());
     }
 
     audiounit_convert_channel_layout(layout.as_ref())
+}
+
+fn get_channel_layout(output_unit: AudioUnit) -> Result<Vec<mixer::Channel>> {
+    audiounit_get_current_channel_layout(output_unit)
+        .or_else(|_| {
+            // The kAudioUnitProperty_AudioChannelLayout property isn't known before
+            // macOS 10.12, attempt another method.
+            cubeb_log!(
+                "Cannot get current channel layout for audiounit @ {:p}. Trying preferred channel layout.",
+                output_unit
+            );
+            audiounit_get_preferred_channel_layout(output_unit)
+        })
 }
 
 fn start_audiounit(unit: AudioUnit) -> Result<()> {
@@ -3050,20 +3061,22 @@ impl<'ctx> CoreStreamData<'ctx> {
                 e
             })?;
 
-            // XXX
-            let device_layout = if using_voice_processing_unit {
-                let mut channels = Vec::with_capacity(output_hw_desc.mChannelsPerFrame as usize);
-                match output_hw_desc.mChannelsPerFrame {
-                    1 => channels.push(mixer::Channel::FrontCenter),
-                    _ => {
-                        channels.push(mixer::Channel::FrontLeft);
-                        channels.push(mixer::Channel::FrontRight);
-                    }
-                }
-                channels
-            } else {
-                audiounit_get_current_channel_layout(self.output_unit)
-            };
+            let device_layout = self
+                .get_output_channel_layout()
+                .map_err(|e| {
+                    cubeb_log!(
+                        "({:p}) Could not get any channel layout. Defaulting to no channels.",
+                        self.stm_ptr
+                    );
+                    e
+                })
+                .unwrap_or_default();
+
+            cubeb_log!(
+                "({:p} Using output device channel layout {:?}",
+                self.stm_ptr,
+                device_layout
+            );
 
             // The mixer will be set up when
             // 1. using aggregate device whose input device has output channels
@@ -3603,6 +3616,28 @@ impl<'ctx> CoreStreamData<'ctx> {
         }
 
         Ok(())
+    }
+
+    fn get_output_channel_layout(&self) -> Result<Vec<mixer::Channel>> {
+        self.debug_assert_is_on_stream_queue();
+        assert!(!self.output_unit.is_null());
+        if !self.using_voice_processing_unit() {
+            return get_channel_layout(self.output_unit);
+        }
+
+        // The VoiceProcessingIO unit (as tried on MacOS 14) is known to not support
+        // kAudioUnitProperty_AudioChannelLayout queries, and to lie about
+        // kAudioDevicePropertyPreferredChannelLayout. If we're using
+        // VoiceProcessingIO, try standing up a regular AudioUnit and query that.
+        cubeb_log!(
+            "({:p}) get_output_channel_layout with a VoiceProcessingIO output unit. Trying a dedicated unit.",
+            self.stm_ptr
+        );
+        let mut dedicated_unit = create_audiounit(&self.output_device)?;
+        let res = get_channel_layout(dedicated_unit);
+        dispose_audio_unit(dedicated_unit);
+        dedicated_unit = ptr::null_mut();
+        res
     }
 }
 
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -2124,9 +2124,13 @@ impl ContextOps for AudioUnitContext {
         state_callback: ffi::cubeb_state_callback,
         user_ptr: *mut c_void,
     ) -> Result<Stream> {
-        if (!input_device.is_null() && input_stream_params.is_none())
-            || (!output_device.is_null() && output_stream_params.is_none())
-        {
+        if !input_device.is_null() && input_stream_params.is_none() {
+            cubeb_log!("Cannot init an input device without input stream params");
+            return Err(Error::invalid_parameter());
+        }
+
+        if !output_device.is_null() && output_stream_params.is_none() {
+            cubeb_log!("Cannot init an output device without output stream params");
             return Err(Error::invalid_parameter());
         }
 
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -971,36 +971,41 @@ fn create_audiounit(device: &device_info) -> Result<AudioUnit> {
 
     if device.flags.contains(device_flags::DEV_INPUT) {
         // Input only.
-        enable_audiounit_scope(unit, DeviceType::INPUT, true).map_err(|e| {
-            cubeb_log!("Fail to enable audiounit input scope. Error: {}", e);
-            Error::error()
-        })?;
-        enable_audiounit_scope(unit, DeviceType::OUTPUT, false).map_err(|e| {
-            cubeb_log!("Fail to disable audiounit output scope. Error: {}", e);
-            Error::error()
-        })?;
+        if let Err(e) = enable_audiounit_scope(unit, DeviceType::INPUT, true) {
+            cubeb_log!("Failed to enable audiounit input scope. Error: {}", e);
+            dispose_audio_unit(unit);
+            return Err(Error::error());
+        }
+        if let Err(e) = enable_audiounit_scope(unit, DeviceType::OUTPUT, false) {
+            cubeb_log!("Failed to disable audiounit output scope. Error: {}", e);
+            dispose_audio_unit(unit);
+            return Err(Error::error());
+        }
     }
 
     if device.flags.contains(device_flags::DEV_OUTPUT) {
         // Output only.
-        enable_audiounit_scope(unit, DeviceType::OUTPUT, true).map_err(|e| {
-            cubeb_log!("Fail to enable audiounit output scope. Error: {}", e);
-            Error::error()
-        })?;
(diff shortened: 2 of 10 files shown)


src/backend/mod.rs
```diff
@@ -2760,7 +2860,20 @@ impl<'ctx> CoreStreamData<'ctx> {
                 e
             })?;
 
-            let device_layout = audiounit_get_current_channel_layout(self.unit);
+            // XXX
```
padenot: Fine for now.

src/backend/mod.rs
```diff
@@ -2421,17 +2421,12 @@ impl<'ctx> CoreStreamData<'ctx> {
     }
 
     fn should_use_aggregate_device(&self) -> bool {
-        // It's impossible to create an aggregate device from an aggregate device, and it's
```
padenot, line 2452: This I want to try again on an older macOS version. I remember having a problem and adding this code.

src/backend/tests/utils.rs
```diff
@@ -885,7 +885,7 @@ impl TestDevicePlugger {
             CFRelease(device_uid as *const c_void);
 
             // This device is private to the process creating it.
-            let private_value: i32 = 1;
```
ChunMinChang: Is the created aggregate device exposed to another process? Why is this necessary?

src/backend/mod.rs
```diff
@@ -2517,38 +2573,110 @@ impl<'ctx> CoreStreamData<'ctx> {
...
-                    in_dev_info.id = device.get_device_id();
-                    out_dev_info.id = device.get_device_id();
-                    in_dev_info.flags = device_flags::DEV_INPUT;
-                    out_dev_info.flags = device_flags::DEV_OUTPUT;
+        let mut using_voice_processing_unit = self.has_input()
+            && self.has_output()
+            && self
+                .input_stream_params
+                .prefs()
+                .contains(StreamPrefs::VOICE);
```
ChunMinChang: Do we have any test that uses `StreamPrefs::VOICE`? A *VoiceProcessingIO* test is necessary for this PR.

If the input device's `AudioDeviceID` is the same as the output device's, is the *VoiceProcessingIO* still needed? (Some Plantronics headset with mic gives the same input and output `AudioDeviceID`)

src/backend/mod.rs
```diff
@@ -1052,13 +1104,10 @@ fn set_device_to_audiounit(
...
 }
 
-fn create_blank_audiounit() -> Result<AudioUnit> {
+fn create_typed_audiounit(sub_type: ::std::os::raw::c_uint) -> Result<AudioUnit> {
     let desc = AudioComponentDescription {
         componentType: kAudioUnitType_Output,
-        #[cfg(not(target_os = "ios"))]
-        componentSubType: kAudioUnitSubType_HALOutput,
-        #[cfg(target_os = "ios")]
-        componentSubType: kAudioUnitSubType_RemoteIO,
```
ChunMinChang, line 1066: This definitely looks better! Thanks!

src/backend/mod.rs
```diff
@@ -1052,13 +1104,10 @@ fn set_device_to_audiounit(
     }
 }
 
-fn create_blank_audiounit() -> Result<AudioUnit> {
+fn create_typed_audiounit(sub_type: ::std::os::raw::c_uint) -> Result<AudioUnit> {
```
ChunMinChang: `c_uint` can be used here with changing `use std::os::raw::c_void` to `std::os::raw::{c_void, c_uint}` in the beginning lines of the file.

src/backend/mod.rs
```diff
@@ -2679,7 +2809,11 @@ impl<'ctx> CoreStreamData<'ctx> {
                 self.input_stream_params.format(),
                 SAFE_MAX_LATENCY_FRAMES as usize,
                 self.input_dev_desc.mChannelsPerFrame as usize,
-                (self.input_dev_desc.mChannelsPerFrame - device_channel_count) as usize,
+                if self.input_dev_desc.mChannelsPerFrame > device_channel_count {
+                    (self.input_dev_desc.mChannelsPerFrame - device_channel_count) as usize
+                } else {
+                    0
+                },
```
```suggestion
self.input_dev_desc.mChannelsPerFrame.saturating_sub(device_channel_count)
```
Nice catch! [`saturating_sub`](https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_sub) is there to do this job.

src/backend/mod.rs
```diff
@@ -2787,7 +2919,20 @@ impl<'ctx> CoreStreamData<'ctx> {
...
 
-            let device_layout = audiounit_get_current_channel_layout(self.output_unit);
+            // XXX
+            let device_layout = if using_voice_processing_unit {
+                let mut channels = Vec::with_capacity(output_hw_desc.mChannelsPerFrame as usize);
+                match output_hw_desc.mChannelsPerFrame {
+                    1 => channels.push(mixer::Channel::FrontCenter),
+                    _ => {
+                        channels.push(mixer::Channel::FrontLeft);
+                        channels.push(mixer::Channel::FrontRight);
```
ChunMinChang: The `mixer::Channel::Silence` should be appended for more than two channels. But what the channel order should be?

src/backend/mod.rs
```diff
@@ -2517,38 +2574,110 @@ impl<'ctx> CoreStreamData<'ctx> {
...
-                    in_dev_info.id = device.get_device_id();
-                    out_dev_info.id = device.get_device_id();
-                    in_dev_info.flags = device_flags::DEV_INPUT;
-                    out_dev_info.flags = device_flags::DEV_OUTPUT;
+        let mut using_voice_processing_unit = self.has_input()
+            && self.has_output()
+            && self
+                .input_stream_params
+                .prefs()
+                .contains(StreamPrefs::VOICE);
```
Pehrsons: I'm thinking we should do this regardless of `StreamPref::VOICE`. I'm not noticing *any* added latency (with nightly and release running side by side) in bypass mode. Thoughts?

src/backend/mod.rs
```diff
@@ -2517,38 +2574,110 @@ impl<'ctx> CoreStreamData<'ctx> {
...
+            let output_unit = if self.has_output() {
+                create_audiounit(&out_dev_info).map_err(|e| {
+                    cubeb_log!("({:p}) Failed to create regular AudioUnit for output.", self.stm_ptr);
+                    e
+                })
+            } else {
+                Ok(ptr::null_mut())
+            }?;
+            Ok::<(AudioUnit, AudioUnit), Error>((input_unit, output_unit))
+        })?;
```
ChunMinChang: The code is hard to read.`setup` is complex enough (although we never find time to fix `clippy::cognitive_complexity`), and the `or_else` pattern brings some unhelpful `Err`s that increases the reading difficulty. To simplify the code, `if let` is a more practical choice. 

On the other hand, with the changes made here, `should_use_aggregate_device` should be removed. Even though `should_use_aggregate_device` returns `true` and devices are eligible to create an aggregate device, the VoiceProcessIO can be used instead. Or the function needs to be renamed, at least.

Using a function for AudioUnit-creation makes `setup` more readable, something like this:

```rust
fn create_audiounits(&mut self) -> Result<(device_info, device_info)> {
    // Create an AudioUnit:
    // - If we're eligible to use voice processing, try creating a VoiceProcessingIO AudioUnit.
    // - If we should use an aggregate device, try creating one and input and output AudioUnits next.
    // - As last resort, create regular AudioUnits. This is also the normal non-duplex path.

    let should_use_voice_processing_unit = self.has_input()
        && self.has_output()
        && self
            .input_stream_params
            .prefs()
            .contains(StreamPrefs::VOICE);

    let should_use_aggregate_device = {
        // ..... bring code here
        // Only use an aggregate device when the device are different.
        self.has_input()
            && self.has_output()
            && self.input_device.id != self.output_device.id
            && !either_already_aggregate
    };

    if should_use_voice_processing_unit {
        if let Ok(au) = create_voiceprocessing_audiounit(&self.input_device, &self.output_device) {
            self.input_unit = au;
            self.output_unit = au;
            return Ok((self.input_device.clone(), self.output_device.clone()));
        }
        // ... error logs here
    }

    if should_use_aggregate_device {
        if let Ok(device) = AggregateDevice::new(self.input_device.id, self.output_device.id) {
            let agg_in_dev_info = device_info {
                id: device.get_device_id(),
                ..self.input_device
            };
            let agg_out_dev_info = device_info {
                id: device.get_device_id(),
                ..self.output_device
            };
            if let (Ok(in_au), Ok(out_au)) = (
                create_audiounit(&agg_in_dev_info),
                create_audiounit(&agg_out_dev_info),
            ) {
                self.input_unit = in_au;
                self.output_unit = out_au;
                self.aggregate_device = Some(device);
                return Ok((agg_in_dev_info, agg_out_dev_info));
            }
        }
        // ... error logs here
    }

    // Regular AudioUnit creation here.
    self.input_unit = if self.has_input() {
        create_audiounit(&self.input_device).map_err(|e| {
            // ... error logs here
            cubeb_log!(
                "({:p}) Failed to create regular AudioUnit for input.",
                self.stm_ptr
            );
            e
        })?
    } else {
        ptr::null_mut()
    };

    self.output_unit = if self.has_output() {
        create_audiounit(&self.output_device).map_err(|e| {
            cubeb_log!(
                "({:p}) Failed to create regular AudioUnit for output.",
                self.stm_ptr
            );
            e
        })?
    } else {
        ptr::null_mut()
    };

    Ok((self.input_device.clone(), self.output_device.clone()))
}
```

so the readers don't need to know the details in the `setup`:

```rust
let (in_dev_info, out_dev_info) = self.create_audiounits()?;
let using_voice_processing_unit = self.input_unit == self.output_unit;
```

src/backend/mod.rs
```diff
@@ -2517,38 +2573,110 @@ impl<'ctx> CoreStreamData<'ctx> {
...
+        })
+        .or_else(|_| {
+            let input_unit = if self.has_input() {
+                create_audiounit(&in_dev_info).map_err(|e| {
+                    cubeb_log!("({:p}) Failed to create regular AudioUnit for input.", self.stm_ptr);
+                    e
+                })
+            } else {
+                Ok(ptr::null_mut())
+            }?;
```
```suggestion
            let input_unit = if self.has_input() {
                create_audiounit(&in_dev_info).map_err(|e| {
                    cubeb_log!("({:p}) Failed to create regular AudioUnit for input.", self.stm_ptr);
                    e
                })?
            } else {
                ptr::null_mut()
            };
```

The `?` can be put after `create_audiounit` so `Ok` can be removed in the `else` block.

Same for output side.

src/backend/mod.rs
```diff
@@ -2787,7 +2918,55 @@ impl<'ctx> CoreStreamData<'ctx> {
...
+                            "({:p}) Could not get current channel layout for dedicated unit. \
+                             Trying to get preferred channel layout with the dedicated unit.",
+                             self.stm_ptr
+                        );
+                        audiounit_get_preferred_channel_layout(dedicated_unit)
+                    });
+                    audio_unit_uninitialize(dedicated_unit);
+                    dispose_audio_unit(dedicated_unit);
+                    dedicated_unit = ptr::null_mut();
+                    res
```
ChunMinChang: In the voice-processing case, the code tries to find the channel-layout by:
1. `get_current` with `VoiceProcessingIO` AU (always fail)
2. `get_current` with a dummy output AU
3. `get_preferred` with a dummy output AU (always work?)
4.  `get_preferred` with `VoiceProcessingIO`

Why are 2 and 3 prioritized over 4? Shouldn't the code respect the `VoiceProcessingIO` first? If 2 or 3 always work, then 4 never called. Does `VoiceProcessingIO` support `kAudioDevicePropertyPreferredChannelLayout`, or it always fails?

If 4 always fails or 4 should be prioritized, then the following code:

```rs
fn get_channel_layout(output_unit: AudioUnit) -> Result<Vec<mixer::Channel>> {
    audiounit_get_current_channel_layout(output_unit)
        .or_else(|_| audiounit_get_preferred_channel_layout(output_unit))
}

// CoreStreamData's method
fn get_output_channel_layout(&self) -> Result<Vec<mixer::Channel>> {
    get_channel_layout(self.output_unit).or_else(|e| {
        // If it's VoiceProcessingIO case, try standing up a regular AudioUnit and query that.
        if self.input_unit == self.output_unit {
            let mut dedicated_unit = create_audiounit(&self.output_device)?;
            let res = get_channel_layout(dedicated_unit);
            audio_unit_uninitialize(dedicated_unit);
            dispose_audio_unit(dedicated_unit);
            dedicated_unit = ptr::null_mut();
            return res;
        }
        Err(e)
    })
}
```

can be used to simplify the code in `setup`:
```rs
let device_layout = self.get_output_channel_layout().unwrap_or_default();
```

ChunMinChang (changes requested):
Thanks for adding VoiceProcessing support!

Since VoiceProcessing code is used behind a pref, they are not tested yet. It would be great to have some tests for this. 

I see some issues with Blackhole devices, but the tests can be run locally without BlackHole as long as it has built-in or plugged audio devices, although some device-related test settings (in Git Actions) might need to be tweaked. 

Some complicated code added to the `setup` can reduce the readability. It's better to simplify the code to lower the long-term maintenance effort.

Pehrsons:
> Since VoiceProcessing code is used behind a pref

Define pref?

src/backend/tests/utils.rs
Pehrsons: See https://github.com/mozilla/cubeb-coreaudio-rs/pull/185/commits/a6db1da3233a0773ffb09bda3e9ad1892cd2e1fc

Pehrsons: relevant output (without the change linked above):
```
      Running unittests src/lib.rs (target/debug/deps/cubeb_coreaudio-18079467f3b516b6)

running 1 test
Before plugging, default Input device is 120
Create plugger device: 162 for Input
After plugging, default Input device is 120

Input devices
--------------------
 125: AppleUSBAudioEngine:Burr-Brown from TI:USB audio CODEC:3400000:2
	uid: USB audio CODEC
 135: AppleUSBAudioEngine:Microsoft:Microsoft® LifeCam HD-3000:3112000:3
	uid: Microsoft® LifeCam HD-3000
  55: BlackHole2ch_UID
	uid: BlackHole 2ch
 120: BuiltInMicrophoneDevice
	uid: MacBook Pro Microphone
 162: org.mozilla.TestAggregateDevice_1697467741403804000
	uid: TestAggregateDevice_1697467741403804000

Current default Input device is 120
state: started
Stream runs on the device 162 for Input
test backend::tests::device_change::test_destroy_input_stream_after_unplugging_a_default_input_device has been running for over 60 seconds
```

src/backend/mod.rs
Pehrsons: Not here, I'll add some. In gecko we might, but I suppose we don't have virtual devices on mac there.

Does identical device ids guarantee no drift? If so we don't need VPIO for the sake of handling drift. But once we have the API to allow **enabling** voice processing features we'll still need it (and be able to toggle it on the fly).

src/backend/mod.rs
Pehrsons: Ooh!

src/backend/mod.rs
Pehrsons: Yes, well, overtaken by https://github.com/mozilla/cubeb-coreaudio-rs/pull/185/commits/5c801e8006bba9ee857ce112d22e81491fbd82b7.

src/backend/mod.rs
Pehrsons: 4 doesn't fail but IIRC it returns garbage. Locally I've seen `7` for a mono input. `13` for another. I'll look more at this later.

src/backend/mod.rs
Pehrsons: Actually, I'll change to using VPIO regardless of `StreamPrefs::VOICE`, because 1) I see no difference at all in latency, and 2) aggregate devices just doesn't handle drift well (at all it seems, at least if the input device is slower than the output device).
Side effects of this should be limited to: 1) the time it takes to init the VPIO AudioUnit is longer, and 2) audio ducking is active while the VPIO AudioUnit is running. I expect 2) to be a good thing in most cases. It is better than audio distortions due to drift in other cases.
As a bonus, that will fix the testing situation.

src/backend/mod.rs
Pehrsons: In case we are using VPIO, should we just ditch 1 and do 2 and 3? Would be simpler.

src/backend/mod.rs
ChunMinChang: By definition, [`CUBEB_STREAM_PREF_VOICE`](https://github.com/mozilla/cubeb/blob/f1bfefbbd412eaa0fe89f5afb3c11b427dc2446a/include/cubeb/cubeb.h#L238-L244) is mainly used to change the device selection, like [here](https://github.com/mozilla/cubeb/blob/f1bfefbbd412eaa0fe89f5afb3c11b427dc2446a/src/cubeb_wasapi.cpp#L2679-L2681). Different settings may be applied in this case though, e.g., for resampler [on WASAPI](https://github.com/mozilla/cubeb/blob/f1bfefbbd412eaa0fe89f5afb3c11b427dc2446a/src/cubeb_wasapi.cpp#L2622-L2623) (maybe we should apply here as well?). Since using vpio doesn't affect device selection from the user's perspective, I don't have a strong opinion, and I am not aware that Gecko has non-WebRTC cases for duplex stream usage. @padenot, @kinetiknz: do you know if the duplex stream is used for non-WebRTC cases (e.g., WebAudio or dolphin emulator)?

src/backend/mod.rs
ChunMinChang: Using 2 and 3 seems reasonable if 4 returns a garbage value.

ChunMinChang:
> > Since VoiceProcessing code is used behind a pref
> 
> Define pref?

What I meant was that whether the vpio is set is by checking the `cubeb_stream_prefs`'s flag.

src/backend/tests/utils.rs
ChunMinChang: Weird. I should upgrade to MacOS 14 to check this. In that case, can you update the comment: `// This device is private to the process creating it.` as well? IIRC, this comment is for `private_value`.

src/backend/mod.rs
padenot: We can land this now, and I'll have folks re-test latency for us. Firefox is used for non-voice use-case, but it isn't clear if VPIO is suitable for non-voice use-case.

The documentation says "can" and is open-ended and not prescriptive.

src/backend/mod.rs
Pehrsons: I added some handling to dispose created audio units on failure, otherwise it looked good. Thanks for the improvements!

src/backend/mod.rs
Pehrsons: I have reverted this, so should be OK.

src/backend/mod.rs
```diff
@@ -2116,6 +2116,8 @@ impl ContextOps for AudioUnitContext {
     ) -> Result<Stream> {
         if (!input_device.is_null() && input_stream_params.is_none())
             || (!output_device.is_null() && output_stream_params.is_none())
+            || (input_stream_params.is_none() && output_stream_params.is_none())
+            || data_callback.is_none()
```
Pehrsons: A bit of a behavior change here, note.

src/backend/mod.rs
```diff
@@ -2617,7 +2771,14 @@ impl<'ctx> CoreStreamData<'ctx> {
             // channels to the audio callback.
             let params = unsafe {
                 let mut p = *self.input_stream_params.as_ptr();
-                p.channels = input_hw_desc.mChannelsPerFrame;
+                p.channels = if using_voice_processing_unit {
+                    // VPIO's input_hw_desc.mChannelsPerFrame reports an oddly high number of channels
+                    // (we've seen 7 channels reported for a mono webcam for instance). Yet it will
+                    // accept any channel count and only spit out a single mono channel.
+                    1
```
Pehrsons: Another behavior change here in that input will always be mono.

src/backend/mod.rs
```diff
@@ -2783,6 +2944,11 @@ impl<'ctx> CoreStreamData<'ctx> {
             let params = unsafe {
                 let mut p = *self.output_stream_params.as_ptr();
                 p.channels = output_hw_desc.mChannelsPerFrame;
+                if using_voice_processing_unit || self.aggregate_device.is_some() {
+                    // VPIO will always use the sample rate of the input hw for both input and output,
+                    // as reported to us. (We can override it but we cannot improve quality this way).
+                    p.rate = output_hw_desc.mSampleRate as _;
```
Pehrsons, line 3042: Another behavior change here in that the output rate will be limited to the input hw rate (regardless of configuration)

src/backend/mod.rs
```diff
@@ -963,39 +979,96 @@ fn create_audiounit(device: &device_info) -> Result<AudioUnit> {
...
-            cubeb_log!("Fail to enable audiounit input scope. Error: {}", e);
-            Error::error()
-        })?;
-        enable_audiounit_scope(unit, DeviceType::OUTPUT, false).map_err(|e| {
-            cubeb_log!("Fail to disable audiounit output scope. Error: {}", e);
-            Error::error()
-        })?;
+        if let Err(e) = enable_audiounit_scope(unit, DeviceType::INPUT, true) {
+            cubeb_log!("Failed to enable audiounit input scope. Error: {}", e);
+            dispose_audio_unit(unit);
```
ChunMinChang, line 992: Nice catch, thanks!

src/backend/mod.rs
```diff
@@ -963,39 +979,96 @@ fn create_audiounit(device: &device_info) -> Result<AudioUnit> {
...
+    in_device: &device_info,
+    out_device: &device_info,
+) -> Result<AudioUnit> {
+    assert!(in_device.flags.contains(device_flags::DEV_INPUT));
+    assert!(!in_device.flags.contains(device_flags::DEV_OUTPUT));
+    assert!(!out_device.flags.contains(device_flags::DEV_INPUT));
+    assert!(out_device.flags.contains(device_flags::DEV_OUTPUT));
+
+    let unit = create_typed_audiounit(kAudioUnitSubType_VoiceProcessingIO)?;
+
```
ChunMinChang, line 1041: I wonder why `enable_audiounit_scope` is not called.

src/backend/mod.rs
```diff
@@ -2513,42 +2714,18 @@ impl<'ctx> CoreStreamData<'ctx> {
...
-                        e
-                    );
-                }
-            }
-        } else {
-            cubeb_log!("Not using an aggregate device");
-        }
+        let (in_dev_info, out_dev_info) = self.create_audiounits()?;
+        let using_voice_processing_unit =
+            !self.input_unit.is_null() && self.input_unit == self.output_unit;
```
ChunMinChang: We can move this into a function. This check appears in other places like `close` or `start_audiounits`.

src/backend/mod.rs
```diff
@@ -3254,6 +3466,26 @@ impl<'ctx> CoreStreamData<'ctx> {
 
         Ok(())
     }
+
+    fn get_output_channel_layout(&self) -> Result<Vec<mixer::Channel>> {
```
ChunMinChang, line 3612:
```rs
assert!(!self.output_unit.is_null())
```

ChunMinChang (commented):
The patches look good. However, I'd like to identify where the crash in the test failures is and its cause, in case the crash happens in the wild.

src/backend/mod.rs
```diff
@@ -2482,6 +2525,164 @@ impl<'ctx> CoreStreamData<'ctx> {
         input_domain == output_domain
     }
 
+    fn create_audiounits(&mut self) -> Result<(device_info, device_info)> {
+        let should_use_voice_processing_unit = self.has_input() && self.has_output();
```
ChunMinChang, line 2602: what if the input or output is an aggregate device? Does it make a difference?

run_device_tests.sh
```diff
@@ -14,7 +14,10 @@ cargo test test_plug_and_unplug_device -- --ignored --nocapture
 
 cargo test test_register_device_changed_callback_to_check_default_device_changed_input -- --ignored --nocapture
 cargo test test_register_device_changed_callback_to_check_default_device_changed_output -- --ignored --nocapture
-cargo test test_register_device_changed_callback_to_check_default_device_changed_duplex -- --ignored --nocapture
+
+# FIXME: The test will hang if the default input or output is an aggregate device (and VoiceProcessingIO is)
```
ChunMinChang: This can be enabled after excluding `VOICEPROCESSING_AGGREGATE_DEVICE_NAME` in the test devices: https://github.com/mozilla/cubeb-coreaudio-rs/actions/runs/6669698195/job/18129159617#step:6:477

Changing `DeviceFilter` into:

```rs
pub enum DeviceFilter {
    ExcludeCubebAggregateAndVPIO,
    IncludeAll,
}
```

and updating `test_get_all_devices` to:

```rs
...
match filter {
    DeviceFilter::ExcludeCubebAggregateAndVPIO => {
        devices.retain(|&device| {
            if let Ok(uid) = get_device_global_uid(device) {
                let uid = uid.into_string();
                !uid.contains(PRIVATE_AGGREGATE_DEVICE_NAME)
                    && !uid.contains(VOICEPROCESSING_AGGREGATE_DEVICE_NAME)
            } else {
                true
            }
        });
    }
    _ => {}
}
```

makes `test_register_device_changed_callback_to_check_default_device_changed_duplex` work on my local machine and git actions. The reference change is here: https://github.com/mozilla/cubeb-coreaudio-rs/commit/f34253c0114b54c8adfca10471e45e837b84e7f8

src/backend/mod.rs
Pehrsons: then creation of vpio fails and we try to create an aggregate device instead.

run_device_tests.sh
Pehrsons: Woot, thanks! I'll fix.

Pehrsons: I'm not having the same luck. See here the logs, plus a new `println!()` in `test_device_channels_in_scope`.

```
running 1 test
NOTICE: The test will hang if the default input or output is an aggregate device.
We will fix this later.
id: 55, scope: input, channels: 2
id: 120, scope: input, channels: 1
id: 113, scope: input, channels: 0
id: 55, scope: output, channels: 2
id: 120, scope: output, channels: 0
id: 113, scope: output, channels: 2
mod.rs:2205: Use global latency 512 instead of the requested latency 4096.
mod.rs:199: Using the system default device
mod.rs:199: Using the system default device
mod.rs:2543: Input device ID: 120 (aggregate: false)
mod.rs:2557: Output device ID: 120 (aggregate: false)
mod.rs:2579: (0x12780b3d0) Using VoiceProcessingIO AudioUnit
mod.rs:2729: (0x12780b3d0) Initializing input by device info: device_info { id: 120, flags: device_flags(DEV_INPUT | DEV_SELECTED_DEFAULT) }
mod.rs:2741: (0x12780b3d0) Opening input side: rate 48000, channels 1, format Float32LE, layout FRONT_CENTER, prefs (empty), latency in frames 512, voice processing true.
mod.rs:2771: (0x12780b3d0) Input hardware description: AudioStreamBasicDescription { mSampleRate: 48000.0, mFormatID: 1819304813, mFormatFlags: 41, mBytesPerPacket: 4, mFramesPerPacket: 1, mBytesPerFrame: 4, mChannelsPerFrame: 7, mBitsPerChannel: 32, mReserved: 0 }
mod.rs:1229: The buffer frame size of AudioUnit 0x97cc30ae for INPUT is already 512
mod.rs:2891: (0x12780b3d0) Input audiounit init with device 120 successfully.
mod.rs:2901: (0x12780b3d0) Initialize output by device info: device_info { id: 55, flags: device_flags(DEV_OUTPUT | DEV_SELECTED_DEFAULT) }
mod.rs:2907: (0x12780b3d0) Opening output side: rate 44100, channels 2, format Float32LE, layout FRONT_LEFT | FRONT_RIGHT, prefs (empty), latency in frames 512, voice processing true.
mod.rs:2937: (0x12780b3d0) Output hardware description: AudioStreamBasicDescription { mSampleRate: 48000.0, mFormatID: 1819304813, mFormatFlags: 9, mBytesPerPacket: 8, mFramesPerPacket: 1, mBytesPerFrame: 8, mChannelsPerFrame: 2, mBitsPerChannel: 32, mReserved: 0 }
mod.rs:3479: (0x12780b3d0) get_output_channel_layout with a VoiceProcessingIO output unit. Trying a dedicated unit.
mod.rs:2986: (0x12780b3d0 Using output device channel layout [FrontLeft, FrontRight]
mod.rs:1229: The buffer frame size of AudioUnit 0x97cc30ae for OUTPUT is already 512
mod.rs:3077: (0x12780b3d0) Output audiounit init with device 55 successfully.
mod.rs:2266: (0x12780b3d0) Cubeb stream init successful.

Input devices
--------------------
id: 55, scope: input, channels: 4
id: 120, scope: input, channels: 3
id: 113, scope: input, channels: 6
  55: BlackHole2ch_UID
	uid: BlackHole 2ch
 120: BuiltInMicrophoneDevice
	uid: MacBook Pro Microphone
 113: BuiltInSpeakerDevice
	uid: MacBook Pro Speakers

Switch device for Input: 120 -> 113
id: 113, scope: input, channels: 6
test backend::tests::device_change::test_register_device_changed_callback_to_check_default_device_changed_duplex has been running for over 60 seconds
```

I can see it working in CI since there's only a BlackHole device, but you say it works locally for you?
It's weird that it reports 6 input channels for an output device but yeah 🤷 
I'll see if I can find a workaround.

Pehrsons: Ok, I have a fix based on [this](https://searchfox.org/wubkat/rev/67be1e0883407c894c71a4c903e5030059a7b5e9/Source/WebCore/platform/mediastream/mac/CoreAudioCaptureDeviceManager.cpp#117).

src/backend/mod.rs
Pehrsons: Same [in webkit](https://searchfox.org/wubkat/rev/67be1e0883407c894c71a4c903e5030059a7b5e9/Source/WebCore/platform/mediastream/mac/CoreAudioSharedUnit.cpp#297-315). Both scopes are enabled by default I guess.

src/backend/mod.rs
padenot: This is fine. It's also the case on e.g. some android and windows.

padenot (approved):
Let's try this -- but we might need to revert back to "regular" audiounit when _not_ doing using the stream for voice if latency is increased.

padenot:
Stealing the review from @ChunMinChang who is busy, this has been reworked significantly and there are green tests now (locally and on CI).

src/backend/mod.rs
padenot: We'll see about this later.

src/backend/mod.rs
padenot: I had missed the question. Web Audio API can use non-voice (e.g. in a musical context, to record an instrument or something).  But we need to be prepare to switch back to non vpio if latency isn't as good, and this is fine.

PR #186, opened 2023-11-10
Monitor explicitly chosen output devices for liveness

Pehrsons:
This mirrors what is done for input devices, and enables the output tests. This fixes cubeb's behavior with Firefox when unplugging a device that is being played to with setSinkId and that does not affect the default device (we used to always install the default-output listener and therefore changes to the default device would trigger a reinit, which would result in an error if the device no longer exists).


Pehrsons:
I see this is a new take on #170, well let me know what you think.

padenot (approved):

PR #187, opened 2023-11-10
Refactor devicechange tests to remove the need for sleeping

Pehrsons:


padenot (approved):
As mentioned elsewhere, this could use a `Channel` but it doesn't matter much.

PR #191, opened 2023-11-20
Remove output channel count restriction

Pehrsons:


padenot (approved):
Mind explaining why we're doing this briefly in the commit message (because that what is expected from cubeb's API as tested in the `cubeb` repo).

PR #192, opened 2023-11-21
Destroying a context with collection change callbacks not removed should crash

padenot:
https://github.com/mozilla/cubeb/blob/54217bca3f3e0cd53c073690a23dd25d83557909/test/test_duplex.cpp#L204-L223 tests this.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -2334,6 +2334,14 @@ impl ContextOps for AudioUnitContext {
 
 impl Drop for AudioUnitContext {
     fn drop(&mut self) {
+        if cfg!(debug_assertions) {
+            let devices = self.devices.lock().unwrap();
+            assert!(
+                devices.input.changed_callback.is_none()
+                    && devices.output.changed_callback.is_none()
+            );
+        }
+
         {
             let controller = self.latency_controller.lock().unwrap();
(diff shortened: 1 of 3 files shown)


src/backend/asdasd.js
```diff
@@ -0,0 +1,8 @@
+let config = { type: "webrtc", "video": {
+      "contentType": "video/avc",
+      "width": 1,
+      "height": 2,
+      "bitrate": 3,
+      "framerate": 4
+    }};
+navigator.mediaCapabilities.decodingInfo(config).then((r) => console.log(r.powerEfficient));
```
Pehrsons: the what

src/backend/mod.rs
```diff
@@ -2346,15 +2354,6 @@ impl Drop for AudioUnitContext {
                 );
             }
         }
-
-        // Make sure all the pending (device-collection-changed-callback) tasks
-        // in queue are done, and cancel all the tasks appended after `drop` is executed.
-        let queue = self.serial_queue.clone();
-        queue.run_final(|| {
```
Pehrsons: I think we still want to wait for pending tasks (or you run the risk of UAFs?) and cancel future tasks, i.e. we should still do an empty run_final().

Pehrsons (changes requested):

Pehrsons (approved):
LGTM

PR #195, opened 2023-11-23
Don't duck the VPIO output device

Pehrsons:
With VPIO, all output streams other than the one of the VPIO AudioUnit itself, on the VPIO output device, are ducked, i.e., their volume is statically reduced.

From manual testing it seems creation of the VPIO AudioUnit causes ducking of all output streams across all devices. After setting the output device on the VPIO AudioUnit, ducking is undone on all output streams except for those on the VPIO output device. The volume ramp-up of 1 second seems close to the ramp-up that happens automatically on other output devices.

With no control surface to control ducking, manually undo it on the VPIO output device for now.

### DIFF
--- a/coreaudio-sys-utils/src/audio_device_extensions.rs
+++ b/coreaudio-sys-utils/src/audio_device_extensions.rs
@@ -1,0 +1,22 @@
+use coreaudio_sys::*;
+
+// See https://opensource.apple.com/source/WebCore/WebCore-7604.5.6/platform/spi/cf/CoreAudioSPI.h.auto.html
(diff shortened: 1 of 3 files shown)


coreaudio-sys-utils/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "coreaudio-sys-utils"
-version = "0.1.0"
+version = "0.1.1"
```
Pehrsons: does it make sense to bump the version?

padenot (approved):
r+, no need to bump the version.

PR #204, opened 2024-01-17
Implement input muting and processing

Pehrsons:
This PR implements the cubeb_stream_set_input_mute and cubeb_stream_set_input_processing_params functions, as well as the associated cubeb_get_supported_input_processing_params.

Input processing through VPIO as implemented supports AEC and NS in combination only.


Pehrsons:
Tested on MacOS 14.0 so far. I'll follow up with some changes as needed by MacOS 12 once I have tested there.

padenot (approved):

PR #210, opened 2024-03-08
Use a context-global shared VPIO unit

Pehrsons:
The shared VPIO unit gets created on the first creation of a duplex stream. It will be used by any request for a duplex VOICE stream, if allowed and possible given the requested devices. With this PR there cannot be two VPIO units for a given context.


padenot (approved):
As discussed on chat.m.o, this could probably be streamlined, it does the job.

PR #215, opened 2024-03-27
Sturdy up VPIO usage with a force-list, allowing it on input-only streams, and not restricting to one instance per context

Pehrsons:
Note that this work is based on the proposed solution in [this doc](https://docs.google.com/document/d/10ClrAs341dSEMghpccu74CW2PxdsEXA4WpuntX-ZZRc/edit?usp=sharing).


padenot (approved):
r+, as discussed.

PR #217, opened 2024-04-16
Add audio dumping, make audio output with professional soundcard work

padenot:
Disregard the build error, I need cubeb-rs and cubeb upstream things to merge first. The first three commits are rebased + changed commit message but are essentially merged commits from @Pehrsons (please double-check regardless). We might want to drop the `test_tone.rs` change from this series.

`misc` is not to be reviewed, and I'll drop it.


Pehrsons (changes requested):
Also [DEBUG: switch test_tone to output 11 channels](https://github.com/mozilla/cubeb-coreaudio-rs/pull/217/commits/fca86f5d1a1db24c3d85e6af47797a73de0fcbd6) should not be landed. Please fix the stack and I'll look again.

padenot:
Yes, sorry. I need other series to merge before this one can run correctly.

src/backend/mod.rs
```diff
@@ -87,6 +91,19 @@ lazy_static! {
     };
 }
 
+cfg_if! {
```
Pehrsons: Did I read the docs right in that `cfg_if` is only needed when you want `else` blocks in your cfg'd code?

src/backend/mod.rs
```diff
@@ -524,6 +546,17 @@ extern "C" fn audiounit_input_callback(
...
+
+            cfg_if! {
+            if #[cfg(feature = "audio_dump")] {
+                        dump_audio(
+                            stm.core_stream_data.audio_dump_input,
+                            input_buffer_list.mBuffers[0].mData,
+                            input_frames * stm.core_stream_data.input_dev_desc.mChannelsPerFrame,
+                        );
+                    }
+                }
```
Pehrsons: Indentation seems weird. (throughout)

src/backend/tests/manual.rs
```diff
@@ -1,6 +1,6 @@
 use super::utils::{
-    test_get_devices_in_scope, test_ops_context_operation, test_ops_stream_operation, Scope,
-    StreamType, TestDeviceInfo, TestDeviceSwitcher,
+    test_disable_log, test_enable_log, test_get_devices_in_scope, test_ops_context_operation,
+    test_ops_stream_operation, Scope, StreamType, TestDeviceInfo, TestDeviceSwitcher,
```
Pehrsons: Either skip these entirely, or put the call to `test_enable_log` behind a user choice, since the terminal will be spammed with logging otherwise, and the terminal is where users interact with the tester program.

Pehrsons (changes requested):
In the commit `Add a test for enabling logging` it was not meant for the logging stuff to land. But if we want it could we only compile log.c in a dev or test environment? Do also note @ChunMinChang is the original author of that bit.

Other comments inline.

src/backend/tests/manual.rs
padenot: Skipped

src/backend/mod.rs
```diff
@@ -708,7 +735,15 @@ extern "C" fn audiounit_output_callback(
...
         audiounit_make_silent(&buffers[0]);
-        return NO_ERR;
+        #[cfg(feature = "audio_dump")]
+        {
+            dump_audio(
+                stm.core_stream_data.audio_dump_output,
+                buffers[0].mData,
+                output_frames * stm.core_stream_data.output_dev_desc.mChannelsPerFrame,
+            );
+            return NO_ERR;
```
Pehrsons: this line needs to go outside the cfg block

src/backend/tests/api.rs
```diff
@@ -376,7 +376,6 @@ fn test_get_default_device_id_with_inout_type() {
 #[test]
 fn test_convert_channel_layout() {
     let pairs = [
-        (vec![kAudioObjectUnknown], vec![mixer::Channel::Silence]),
```
Pehrsons, line 379: Don't we want to keep this case but with `s/Silence/Discrete` ?

Pehrsons (approved):
Thanks! One issue and a nit really, fix at your own discretion.

src/backend/tests/api.rs
padenot: `kAudioObjectUnknown` vs. `kAudioChannelLabel_Unknown`. If the former happens, I want to know about it.

padenot: (using the `eprintln!(...)`)

PR #222, opened 2024-05-27
When playing stereo, address the channels by name rather than using 0 and 1

padenot:
This should fix https://bugzilla.mozilla.org/show_bug.cgi?id=1474175.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -3906,26 +3906,53 @@ impl<'ctx> CoreStreamData<'ctx> {
                 device_layout
             );
 
-            // The mixer will be set up when
-            // 1. using aggregate device whose input device has output channels
-            // 2. output device has more channels than we need
-            // 3. output device has different layout than the one we have
-            self.mixer = if self.output_dev_desc.mChannelsPerFrame
-                != self.output_stream_params.channels()
-                || device_layout != mixer::get_channel_order(self.output_stream_params.layout())
+            // Simple case of stereo output only, map to the stereo pair (that might not be the first two channels)
+            if !self.has_input()
+                && self.output_stream_params.channels() == 2
+                && self.output_stream_params.layout() == ChannelLayout::STEREO
             {
-                cubeb_log!("Incompatible channel layouts detected, setting up remixer");
-                // We will be remixing the data before it reaches the output device.
-                Some(Mixer::new(
-                    self.output_stream_params.format(),
-                    self.output_stream_params.channels() as usize,
-                    self.output_stream_params.layout(),
-                    self.output_dev_desc.mChannelsPerFrame as usize,
-                    device_layout,
-                ))
+                let layout = AudioChannelLayout {
+                    mChannelLayoutTag: kAudioChannelLayoutTag_Stereo,
+                    ..Default::default()
+                };
+                let r = audio_unit_set_property(
+                    self.output_unit,
+                    kAudioUnitProperty_AudioChannelLayout,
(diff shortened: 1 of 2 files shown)


src/backend/mod.rs
```diff
@@ -3906,26 +3906,56 @@ impl<'ctx> CoreStreamData<'ctx> {
...
-            // The mixer will be set up when
-            // 1. using aggregate device whose input device has output channels
-            // 2. output device has more channels than we need
-            // 3. output device has different layout than the one we have
-            self.mixer = if self.output_dev_desc.mChannelsPerFrame
-                != self.output_stream_params.channels()
-                || device_layout != mixer::get_channel_order(self.output_stream_params.layout())
+            // Simple case of mono or stereo output only, map to the stereo pair (that might not be the first two channels)
+            if !self.has_input()
+                && self.output_stream_params.channels() == 2
```
Pehrsons, line 3911: This doesn't cover mono.

Pehrsons (changes requested):

src/backend/mod.rs
```diff
@@ -3906,26 +3906,56 @@ impl<'ctx> CoreStreamData<'ctx> {
...
-                    self.output_stream_params.channels() as usize,
-                    self.output_stream_params.layout(),
-                    self.output_dev_desc.mChannelsPerFrame as usize,
-                    device_layout,
-                ))
+                let mut layout = AudioChannelLayout::default();
+                if self.output_stream_params.channels() == 2 {
+                    layout.mChannelLayoutTag = kAudioChannelLayoutTag_Stereo;
+                } else {
+                    layout.mChannelLayoutTag = kAudioChannelLayoutTag_Mono;
```
Pehrsons: Still, this will never get hit.

src/backend/tests/manual.rs
```diff
@@ -766,3 +767,110 @@ fn test_stream_tester() {
         }
     }
 }
+
+// Simple stereo tone test
+#[ignore]
+#[test]
+fn test_tone() {
```
Pehrsons, line 773: We might be able to test this with `blackhole-16ch` by setting `kAudioDevicePropertyPreferredChannelsForStereo` to something that is not `[0, 1]`.
Maybe even on `blackhole-2ch` by just flipping left and right, i.e. setting that property to `[1, 0]`.

Pehrsons (approved):
This kind of thing seems very likely to go undetected if regressed. I'd love to have a test case, see an idea inline.
I'll leave it to your own discretion, r+.

src/backend/tests/manual.rs
padenot: Turns out this isn't possible unfortunately, so we'll have to rely on manual testing for this.

uberjay:
For someone totally unfamiliar with the Firefox release process -- when might this fix make its way into something I can try?

Pehrsons:
It will be in Nightly soon. Follow along in https://bugzilla.mozilla.org/show_bug.cgi?id=1474175.

padenot:
@uberjay this should be available tomorrow in [Nightly](https://www.mozilla.org/fr/firefox/channel/desktop/#nightly). This installs alongside a regular Firefox fine without touching your main profile if it's just for testing (not sure how familiar you are with nightly).

General availability in Firefox release is going to be in Firefox 128, July 8th this year.

uberjay:
Fantastic, the nighty build I just tried plays using the correct channels!!! Thank you all so much, it's greatly appreciated! ❤️

PR #231, opened 2024-08-06
Using `map_err` over `inspect_err`

ChunMinChang:
This patch fixes clippy::manual-inspect error. See [1] for details.

[1] https://rust-lang.github.io/rust-clippy/master/index.html#/manual_inspect


Pehrsons (approved):
Maybe log the `e`s instead of remove them? Seems useful.

LGTM either way

ChunMinChang:
> Maybe log the `e`s instead of remove them? Seems useful.
> 
> LGTM either way

Thanks for the review!

Those errors are lib libcubeb's errors like `InvalidFormat` or `InvalidParameter,` not providing helpful information for debugging, so I didn't log them. I'll leave them as they are for now.

PR #233, opened 2024-08-07
Use different scopes on the output bus depending on AudioUnit type

Pehrsons:
As the inline comment says it seems a HAL unit for a device with >2 output channels reports only 2 channels on the input scope of the output bus. Not sure what these 2 reflect but let's differentiate VPIO (where input scope reports 1 which we want, and output scope reflects hw) and HAL units when selecting scope as they're the only types we use.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -3783,7 +3783,16 @@ impl<'ctx> CoreStreamData<'ctx> {
             let r = audio_unit_get_property(
                 self.output_unit,
                 kAudioUnitProperty_StreamFormat,
-                kAudioUnitScope_Input,
+                if using_voice_processing_unit {
+                    // With a VPIO unit the output scope includes all channels in the hw.
+                    // The VPIO unit however is only MONO which the input scope reflects.
+                    kAudioUnitScope_Input
+                } else {
+                    // With a HAL unit the output scope for the output bus returns the number of
+                    // output channels of the hw, as we want. The input scope seems limited to
+                    // two channels.
+                    kAudioUnitScope_Output
+                },
                 AU_OUT_BUS,
                 &mut output_hw_desc,
                 &mut size,
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -3819,13 +3819,44 @@ impl<'ctx> CoreStreamData<'ctx> {
                 return Err(Error::error());
             }
 
+            // Simple case of stereo output, map to the stereo pair (that might not be the first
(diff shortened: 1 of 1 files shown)


ChunMinChang (commented):
Thanks for the change! 

Are there other coming commits? Landing this individually may cause a noise issue. 

With this patch, the [`mNumberChannels`](https://github.com/mozilla/cubeb-coreaudio-rs/blob/e143e87de1a0521796e10bd454ebfc257c4dcf8c/src/backend/mod.rs#L875) of the given `AudioBuffer` provided by `audiounit_output_callback` will be [matched](https://github.com/mozilla/cubeb-coreaudio-rs/blob/e143e87de1a0521796e10bd454ebfc257c4dcf8c/src/backend/mod.rs#L3928) to the channel count of the device, since now the `output_dev_desc.mChannelsPerFrame` matches `output_hw_desc.mChannelsPerFrame`'s value.

However, if the channel count mismatches the `output_stream_params.channels()` channel, no mixer will be set up, and there will not be enough or too much data for the [`output buffer`](https://github.com/mozilla/cubeb-coreaudio-rs/blob/e143e87de1a0521796e10bd454ebfc257c4dcf8c/src/backend/mod.rs#L784) in the callback since the buffer (prepared by resampler) is for the `output_stream_params.channels()` channel.

I tested this with my [Cubilux Surround Sound Card (CA7)](https://www.cubilux.com/products/7-1-5-1surround-sound-crad) with [test_tone](https://github.com/mozilla/cubeb-coreaudio-rs/blob/e143e87de1a0521796e10bd454ebfc257c4dcf8c/src/backend/tests/manual.rs#L773). Noise appears in all the channels.

Pehrsons:
The new commit updates `output_dev_desc` so we configure the device correctly for 2 channels when deciding to skip the mixer. With my Studio Display that has 8 channels I get an error with the stereo variant of `test_tone` though:
`AudioConverter.cpp:748   Failed to created a new in process converter -> from  2 ch,  48000 Hz, lpcm (0x00000009) 32-bit little-endian float, packed in 16 bytes low-aligned to  8 ch,  48000 Hz, Float32, interleaved, with status 1718449215`
Note `1718449215` is the fourcc for `fmt?`, i.e. [kAudioConverterErr_FormatNotSupported](https://osstatus.com/search/results?platform=all&framework=all&search=1718449215).

Pehrsons:
Ah my bug, fixed now. With this stack playing something >2ch should work on a device with >2ch.

ChunMinChang (approved):
LGTM! thanks!

PR #234, opened 2024-08-08
Delay reinitialization when the stream is paused while the change happens

padenot:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1907780

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -4877,6 +4877,8 @@ impl<'ctx> Drop for AudioUnitStream<'ctx> {
 
 impl<'ctx> StreamOps for AudioUnitStream<'ctx> {
     fn start(&mut self) -> Result<()> {
+        let was_stopped = self.stopped.load(Ordering::SeqCst);
+        let was_draining = self.draining.load(Ordering::SeqCst);
         self.stopped.store(false, Ordering::SeqCst);
         self.draining.store(false, Ordering::SeqCst);
 
@@ -4886,22 +4888,33 @@ impl<'ctx> StreamOps for AudioUnitStream<'ctx> {
                 // Need reinitialization: device was changed when paused. It will be started after
                 // reinit because self.stopped is false.
                 if self.delayed_reinit {
-                    self.reinit().inspect_err(|_| {
+                    let rv = self.reinit().inspect_err(|_| {
                         cubeb_log!(
                             "({:p}) delayed reinit during start failed.",
                             self.core_stream_data.stm_ptr
                         );
-                    })?;
+                    });
+                    // In case of failure, restore the state
+                    if rv.is_err() {
+                        self.stopped.store(was_stopped, Ordering::SeqCst);
+                        self.draining.store(was_draining, Ordering::SeqCst);
+                        return rv;
+                    }
                     self.delayed_reinit = false;
                     Ok(())
                 } else {
                     // Execute start in serial queue to avoid racing with destroy or reinit.
-                    self.core_stream_data.start_audiounits().inspect_err(|_| {
+                    let rv = self.core_stream_data.start_audiounits();
+                    if rv.is_err() {
                         cubeb_log!("({:p}) start failed.", self.core_stream_data.stm_ptr);
-                    })
+                        self.stopped.store(was_stopped, Ordering::SeqCst);
+                        self.draining.store(was_draining, Ordering::SeqCst);
+                        return rv;
+                    }
+                    Ok(())
                 }
             })
-            .unwrap();
+            .unwrap()?;
 
         self.notify_state_changed(State::Started);
 
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -1019,11 +1019,14 @@ extern "C" fn audiounit_property_listener_callback(
     {
         let callback = stm.device_changed_callback.lock().unwrap();
         if let Some(device_changed_callback) = *callback {
+            cubeb_log!("Calling device changed callback");
             unsafe {
                 device_changed_callback(stm.user_ptr);
             }
         }
     }
+
+    cubeb_log!("Reinitializing stream with new device because of device change, async");
     stm.reinit_async();
 
     NO_ERR
@@ -4706,6 +4709,7 @@ impl<'ctx> AudioUnitStream<'ctx> {
                 .flags
                 .contains(device_flags::DEV_SELECTED_DEFAULT)
         {
+            cubeb_log!("Using new default output device");
             self.core_stream_data.output_device =
                 match create_device_info(kAudioObjectUnknown, DeviceType::OUTPUT) {
                     None => {
@@ -4724,6 +4728,7 @@ impl<'ctx> AudioUnitStream<'ctx> {
                 .flags
                 .contains(device_flags::DEV_SELECTED_DEFAULT)
         {
+            cubeb_log!("Using new default input device");
             self.core_stream_data.input_device =
                 match create_device_info(kAudioObjectUnknown, DeviceType::INPUT) {
                     None => {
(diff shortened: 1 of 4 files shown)


src/backend/mod.rs
```diff
@@ -1019,11 +1019,14 @@ extern "C" fn audiounit_property_listener_callback(
...
         let callback = stm.device_changed_callback.lock().unwrap();
         if let Some(device_changed_callback) = *callback {
+            cubeb_log!("Calling device changed callback");
             unsafe {
                 device_changed_callback(stm.user_ptr);
             }
         }
     }
+
+    cubeb_log!("Reinitializing stream with new device, async");
```
Pehrsons: nit: Please say that this is because of a device change.

src/backend/mod.rs
```diff
@@ -4870,14 +4880,25 @@ impl<'ctx> StreamOps for AudioUnitStream<'ctx> {
...
+                        );
+                    })?;
+                    self.delayed_reinit = false;
+                }
+                // Execute start in serial queue to avoid racing with destroy or reinit.
+                self.core_stream_data.start_audiounits().inspect_err(|_| {
+                    cubeb_log!("({:p}) start failed.", self.core_stream_data.stm_ptr);
+                })
+            })
+            .ok_or(Error::error());
```
Pehrsons: I think it's fine to `.unwrap()` here.

src/backend/mod.rs
```diff
@@ -4870,14 +4880,25 @@ impl<'ctx> StreamOps for AudioUnitStream<'ctx> {
...
-        // Execute start in serial queue to avoid racing with destroy or reinit.
-        let result = self
-            .queue
+        self.queue
             .clone()
-            .run_sync(|| self.core_stream_data.start_audiounits())
-            .unwrap();
-
-        result?;
+            .run_sync(|| -> Result<()> {
```
Pehrsons, line 4887: You must return the error if there was one from this `run_sync` block.

src/backend/mod.rs
```diff
@@ -4870,14 +4880,25 @@ impl<'ctx> StreamOps for AudioUnitStream<'ctx> {
...
+                // Need reinitialization: device was changed when paused.
+                if self.delayed_reinit {
+                    self.reinit().inspect_err(|_| {
+                        cubeb_log!(
+                            "({:p}) delayed reinit during start failed.",
+                            self.core_stream_data.stm_ptr
+                        );
+                    })?;
+                    self.delayed_reinit = false;
```
Pehrsons: No need to call `start_audiounits` if `reinit` succeeded. The test should probably check that the stream is in the expected state after starting also.

src/backend/tests/device_change.rs
```diff
@@ -73,6 +75,112 @@ fn test_switch_device_in_scope(scope: Scope) {
...
+                stm.core_stream_data.input_device.id,
+                device_switcher.current()
+            );
+        }
+
+        let after = if scope == Scope::Output {
+            stm.core_stream_data.output_unit
+        } else {
+            stm.core_stream_data.input_unit
+        };
```
Pehrsons: nit: you could shrink the business logic a bit if you inline two closures for getting the current device id and audio unit, respectively, since this pattern repeats.

src/backend/tests/device_change.rs
```diff
@@ -73,6 +75,112 @@ fn test_switch_device_in_scope(scope: Scope) {
...
+                thread::sleep(ten_millis);
+            }
+        }
+
+        // Start the stream, and check that the device in use isn't the same as before pausing
+        assert_eq!(unsafe { OPS.stream_start.unwrap()(stream) }, ffi::CUBEB_OK);
+
+        if scope == Scope::Output {
+            assert_eq!(
+                stm.core_stream_data.output_device.id,
```
Pehrsons: nit: perhaps higher confidence reading the current device id from the unit itself using `kAudioOutputUnitProperty_CurrentDevice`.

Pehrsons (changes requested):
You removed the error early-exit guard on the result from the `run_sync` block in `start`. That needs to be fixed.
I think the test should also check that the audio unit has started, once `start` with the delayed reinit returns success. Either by checking an internal state, or that the state callback fired a started event.
Consider the rest nits.

src/backend/mod.rs
padenot: We should just return an error no?

src/backend/tests/device_change.rs
padenot: I added a two-way comparison so that should now catch any state inconsistency, hopefully.

src/backend/mod.rs
Pehrsons: The only time `run_sync` [returns `None`](https://searchfox.org/mozilla-central/rev/0c55d51c0d2a9b672e42ad40ea54f90267f92a8e/third_party/rust/coreaudio-sys-utils/src/dispatch.rs#157,164-167,174) is if the queue is canceled (`run_final`'s task [has finished](https://searchfox.org/mozilla-central/rev/0c55d51c0d2a9b672e42ad40ea54f90267f92a8e/third_party/rust/coreaudio-sys-utils/src/dispatch.rs#198-200)). That is not supposed to happen here so better that we fail hard.

src/backend/mod.rs
```diff
@@ -4870,14 +4880,28 @@ impl<'ctx> StreamOps for AudioUnitStream<'ctx> {
...
+                // reinit because self.stopped is false.
+                if self.delayed_reinit {
+                    self.reinit().inspect_err(|_| {
+                        cubeb_log!(
+                            "({:p}) delayed reinit during start failed.",
+                            self.core_stream_data.stm_ptr
+                        );
+                    })?;
+                    self.delayed_reinit = false;
+                    Ok(())
```
Pehrsons: I think this is fine when `reinit` fails. `delayed_reinit` remains `true`, but `stopped` has already been set to `false`, but calling `start` again would trigger the reinit again as I'd expect.

There's a latent issue though. If this has happened, `stopped` is `false` but `reinit` failed so the audio units aren't running, then a new device changed event *will start the stream again*, because it looks at `stopped`. Rare, maybe, but a weird failure mode certainly.

src/backend/tests/device_change.rs
```diff
@@ -73,6 +75,127 @@ fn test_switch_device_in_scope(scope: Scope) {
...
+        let stm = unsafe { &mut *(stream as *mut AudioUnitStream) };
+        let before = if scope == Scope::Output {
+            stm.core_stream_data.output_unit
+        } else {
+            stm.core_stream_data.input_unit
+        };
+
+        let check_devices = |current| {
+            stm.queue.run_sync(|| {
+                let bus = if scope == Scope::Output {
```
Pehrsons: nit: use a 3-tuple instead of 3 ifs?

Pehrsons (approved):
Thanks. I see a latent issue if `start` fails. It might get exacerbated by adding `reinit` to a path in `start`.

src/backend/mod.rs
padenot: I restored the state in a new patch.

src/backend/mod.rs
```diff
@@ -4867,17 +4877,39 @@ impl<'ctx> Drop for AudioUnitStream<'ctx> {
...
+                    if rv.is_err() {
+                        self.stopped.store(was_stopped, Ordering::SeqCst);
+                        self.draining.store(was_draining, Ordering::SeqCst);
+                        return Err(Error::error());
+                    }
+                    self.delayed_reinit = false;
+                    Ok(())
+                } else {
+                    // Execute start in serial queue to avoid racing with destroy or reinit.
+                    self.core_stream_data.start_audiounits().inspect_err(|_| {
```
Pehrsons: Do we assume `start_audiounits` doesn't fail in practice, or do we not reset state when it fails in order to not change existing behavior?

Pehrsons (approved):

src/backend/mod.rs
padenot: I mirrored the code here as well, we'll see.

PR #238, opened 2024-10-18
Mark some spammy logs as verbose.

mutexlox-signal:
In a very brief usage of cubeb-coreaudio-rs, I observed substantial log output from the following lines: (format is "&lt;count&gt; &lt;file&gt;:&lt;line&gt;")

 880  mod.rs:1767
 880  mod.rs:1848
1479  mod.rs:3274
1618  mod.rs:1758
2498  mod.rs:1715
3236  mod.rs:1727
8870  device_property.rs:268
8870  device_property.rs:282
9708  mod.rs:2014

Of these, only one (mod.rs:1848) is an error. The others are informational, so make them verbose logs.

### DIFF
--- a/src/backend/device_property.rs
+++ b/src/backend/device_property.rs
@@ -265,7 +265,7 @@ pub fn get_device_streams(
         debug_assert!(devices.contains(&id));
         devices.sort();
         let next_id = devices.into_iter().skip_while(|&i| i != id).nth(1);
-        cubeb_log!(
+        cubeb_logv!(
             "Filtering input streams {:?} for device {}. Next device is {:?}.",
             device_streams
                 .iter()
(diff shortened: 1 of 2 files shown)


mutexlox-signal:
cc @padenot @kinetiknz for review :)

padenot:
This is intentional. Those logging statement are at the level "debug", and those are debugging information. Verbose is when outputting logging messages once per callback.

Generally, when logging in "debug" mode, all the information that let the cubeb developer understand how a stream was created, including the various workarounds on all platforms, are logged. Verbose is for things that are really verbose, happen during capture/rendering. Those aren't strict rules.

mutexlox-signal:
Got it, okay. It might be nice to have a distinction between "info" level logs and "warning/error" level logs, but that sounds like it would be a substantially larger project.

PR #243, opened 2024-11-13
Do not send extra channels for aggregate devices.

mutexlox-signal:
We drop the **first** several channels if there are extras (e.g. for aggregate devices) After f8df86f753d81c3ef2e8a6661b11ed65af579919, this is often incorrect
(because the first device is the input device), and also is inefficient.

So, only request the channels we're planning to use.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -3612,16 +3612,15 @@ impl<'ctx> CoreStreamData<'ctx> {
                 self.stm_ptr,
                 input_hw_desc
             );
-            // Notice: when we are using aggregate device, the input_hw_desc.mChannelsPerFrame is
-            // the total of all the input channel count of the devices added in the aggregate device.
-            // Due to our aggregate device settings, the data captured by the output device's input
-            // channels will be put in the beginning of the raw data given by the input callback.
-
-            // Always request all the input channels of the device, and only pass the correct
-            // channels to the audio callback.
+            // Notice: when we are using aggregate device, input_hw_desc.mChannelsPerFrame is the
+            // total of all the input channel count of the devices added in the aggregate device.
+            // Because we set the input device first on the aggregate device, the input device's
+            // input channels will also be first among all the aggregate device's channels, when
+            // accessed in the input callback. By requesting only the input device's channels here,
+            // any other input channels, i.e. from the output device, will be truncated away.
             let params = unsafe {
                 let mut p = *self.input_stream_params.as_ptr();
-                p.channels = input_hw_desc.mChannelsPerFrame;
+                p.channels = device_channel_count;
                 // Input AudioUnit must be configured with device's sample rate.
                 // we will resample inside input callback.
                 p.rate = input_hw_desc.mSampleRate as _;


mutexlox-signal:
cc @Pehrsons , who I've been working with on this. You may want to do more extensive testing here; I've only tested one combination of input/output devices.

mutexlox-signal:
I believe the test failures are the same as on ToT, but I'm not sure if there are any new problems here.

src/backend/mod.rs
```diff
@@ -3621,7 +3621,6 @@ impl<'ctx> CoreStreamData<'ctx> {
             // channels to the audio callback.
             let params = unsafe {
                 let mut p = *self.input_stream_params.as_ptr();
-                p.channels = input_hw_desc.mChannelsPerFrame;
```
Pehrsons, line 3624: `p.channels` will now be what's requested for the cubeb input stream. But it needs to be the number of hw channels of the requested input device.

src/backend/mod.rs
```diff
@@ -3621,7 +3621,6 @@ impl<'ctx> CoreStreamData<'ctx> {
             // channels to the audio callback.
```
Pehrsons: we should update these comments

Pehrsons (changes requested):

src/backend/mod.rs
mutexlox-signal: Done.

src/backend/mod.rs
mutexlox-signal: Done, thanks for catching this!

src/backend/mod.rs
```diff
@@ -3612,16 +3612,11 @@ impl<'ctx> CoreStreamData<'ctx> {
                 self.stm_ptr,
                 input_hw_desc
             );
-            // Notice: when we are using aggregate device, the input_hw_desc.mChannelsPerFrame is
-            // the total of all the input channel count of the devices added in the aggregate device.
-            // Due to our aggregate device settings, the data captured by the output device's input
-            // channels will be put in the beginning of the raw data given by the input callback.
```
Pehrsons, line 3618: Please keep this comment, but update it to make it accurate: the "our aggregate device settings" refers to the input device coming after the output device, for instance. Rather explain that the input device is set first on the aggregate device, so we only need to request those channels and CoreAudio will drop the rest automatically.

Pehrsons (approved):
Thanks, and sorry for the late review!
LGTM with some work on the comment.

src/backend/mod.rs
mutexlox-signal: Sorry for the delay; I lost track of this and have now forgotten all context. Can you please suggest a specific phrasing, and I can update the PR?

Pehrsons: Perhaps
```
Notice: when we are using aggregate device, input_hw_desc.mChannelsPerFrame is the
total of all the input channel count of the devices added in the aggregate device.
Because we set the input device first on the aggregate device, the input device's
input channels will also be first among all the aggregate device's channels, when
accessed in the input callback. By requesting only the input device's channels here,
any other input channels, i.e. from the output device, will be truncated away.
```

mutexlox-signal: Thanks! Done.

Pehrsons (approved):

PR #245, opened 2024-11-15
Check in Cargo.lock file.

mutexlox-signal:
This makes builds more reproducible.


mutexlox-signal:
cc @Pehrsons @padenot for review on Monday

Issue #250, opened 2025-03-03
`devid` is not stable across disconnection/reconnection


mutexlox-signal:
The `devid` field is not stable on macOS, which I understand might be a requirement on other platforms.

To reproduce, plug in a USB mic (for example), print its dev_id, wait 30 seconds, and then while the code is sleeping unplug the mic and plug it in again. The dev_id will change.

Sample log snippets:

```
<snip>
cubeb.c:670:DeviceID: "BuiltInMicrophoneDevice"
	Name:	"MacBook Pro Microphone"
	Group:	"builtin-internal-mic|spk"
	Vendor:	"Apple Inc."
	Type:	input
	State:	enabled
	Maximum channels:	1
	Format:	S16LE S16BE F32LE F32BE (0x3030) (default: F32LE)
	Rate:	[4
cubeb.c:670:DeviceID: "BlackHole16ch_UID"
	Name:	"BlackHole 16ch"
	Group:	"BlackHole16ch_ModelUID"
	Vendor:	"Existential Audio Inc."
	Type:	input
	State:	enabled
	Maximum channels:	16
	Format:	S16LE S16BE F32LE F32BE (0x3030) (default: F32LE)
	Rate:	[8000
cubeb.c:670:DeviceID: "BlackHole2ch_UID"
	Name:	"BlackHole 2ch"
	Group:	"BlackHole2ch_ModelUID"
	Vendor:	"Existential Audio Inc."
	Type:	input
	State:	enabled
	Maximum channels:	2
	Format:	S16LE S16BE F32LE F32BE (0x3030) (default: F32LE)
	Rate:	[8000, 76
cubeb.c:670:DeviceID: "AppleUSBAudioEngine:Unknown Manufacturer:HD Pro Webcam C920:FFA25D4F:3" (PREFERRED)
	Name:	"HD Pro Webcam C920"
	Group:	"HD Pro Webcam C920:046D:08E5"
	Vendor:	"Unknown Manufacturer"
	Type:	input
	State:	enabled
	Maximum channels:	2
input device: 0x53 Some("BuiltInMicrophoneDevice") Some("MacBook Pro Microphone")
input device: 0x58 Some("BlackHole16ch_UID") Some("BlackHole 16ch")
input device: 0x69 Some("BlackHole2ch_UID") Some("BlackHole 2ch")
[...]

PR #251, opened 2025-04-01
Use a stable identifier for the device when returning a cubeb_devid.

kinetiknz:
Previously we used the AudioObjectID associated with the device, but this value is not stable (e.g. disconnecting and reconnecting the same audio device can result in the AudioObjectID changing).  Instead, use the device's UID after interning it in the cubeb context to provide a stable (for the lifetime of the context) opaque pointer that can be mapped back to the UID and associated AudioObjectID as needed.

Fixes issue #250.

This is similar to the approach we already use for the [PulseAudio](https://github.com/mozilla/cubeb-pulse-rs/blob/6bac666467e4a37cf057f0e17e8c9e8a024b060b/src/backend/context.rs#L352) and [WASAPI](https://github.com/mozilla/cubeb/blob/f8633c0c9217ed5499fdda391946d416c52e6809/src/cubeb_wasapi.cpp#L2411) backends.

### DIFF
--- a/src/backend/device_property.rs
+++ b/src/backend/device_property.rs
@@ -18,6 +18,45 @@ pub fn get_device_uid(
     }
 }
 
+pub fn get_device_from_devid(
+    devid: ffi::cubeb_devid,
+) -> std::result::Result<AudioDeviceID, OSStatus> {
+    debug_assert_running_serially();
+    if devid.is_null() {
+        return Ok(kAudioObjectUnknown);
+    }
+
+    let devid_cstr = unsafe { CStr::from_ptr(devid as _) };
+    let mut devid_cfsr = cfstringref_from_string(devid_cstr.to_str().unwrap());
+
+    let mut device_id: AudioDeviceID = kAudioObjectUnknown;
+
+    let mut translation_value = AudioValueTranslation {
+        mInputData: &mut devid_cfsr as *mut CFStringRef as *mut c_void,
+        mInputDataSize: mem::size_of::<CFStringRef>() as u32,
+        mOutputData: &mut device_id as *mut AudioDeviceID as *mut c_void,
+        mOutputDataSize: mem::size_of::<AudioDeviceID>() as u32,
+    };
+
+    let address = get_property_address(
+        Property::HardwareDeviceForUID,
+        DeviceType::INPUT | DeviceType::OUTPUT,
+    );
+
+    let mut translation_size: usize = mem::size_of::<AudioValueTranslation>();
+    let ret = audio_object_get_property_data(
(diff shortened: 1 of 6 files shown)


src/backend/device_property.rs
```diff
@@ -429,6 +469,7 @@ impl From<Property> for AudioObjectPropertySelector {
             Property::DeviceUID => kAudioDevicePropertyDeviceUID,
             Property::HardwareDefaultInputDevice => kAudioHardwarePropertyDefaultInputDevice,
             Property::HardwareDefaultOutputDevice => kAudioHardwarePropertyDefaultOutputDevice,
+            Property::HardwareDeviceForUID => kAudioHardwarePropertyDeviceForUID,
```
Pehrsons, line 472: Interestingly this one is is `AudioHardwareDeprecated.h`, though it doesn't appear to be officially deprecated?

src/backend/intern.rs
```diff
@@ -0,0 +1,58 @@
+// Copyright © 2017-2018 Mozilla Foundation
```
Pehrsons, line 1: Should this be updated?

src/backend/intern.rs
```diff
@@ -0,0 +1,58 @@
...
+pub struct Intern {
+    vec: Vec<CString>,
+}
+
+impl Intern {
+    pub fn new() -> Intern {
+        Intern { vec: Vec::new() }
+    }
+
+    pub fn add(&mut self, string: &CStr) -> *const c_char {
```
Pehrsons, line 19: Was a bit confused with the naming here, since it doesn't add if it already exists. That said I don't have a suggestion for a better name.

Pehrsons (approved):
I don't have much, LGTM.

src/backend/device_property.rs
kinetiknz: It looks like this moved in macOS 10.8, likely as part of the deprecation of the AudioHardware* APIs [announced for 10.5 -> 10.6](https://developer.apple.com/library/archive/technotes/tn2223/_index.html).  There doesn't seem to be an alternative to use, so I guess this is fine until something changes.

src/backend/intern.rs
kinetiknz: It probably should, but at the moment this file is identical to the copy in cubeb-pulse-rs.

src/backend/intern.rs
kinetiknz: I'll leave it for now (since it's a straight copy from cubeb-pulse-rs), but you're right that this could use a better name.  Maybe `intern()` would be clearer than `add()`.

PR #253, opened 2025-06-20
Fix CI errors

marcin-serwin:


mutexlox-signal:
(As a random user of cubeb with no approval power, thanks for these patches!)

kinetiknz:
Thanks! I've merged the warning fixes and the default input device fix.  I need to take a closer look at the `should_cancel` UAF.

marcin-serwin:
What is the reasoning behind changes in https://github.com/mozilla/cubeb-coreaudio-rs/commit/37fd803332f4e13264ed26a7e81f279cbf2476b8? It seems wrong to me because the lifetime of context is tied to the underlying queue and not the guard. This is especially evident in the `run_async` since there  the guard is usually dropped before the closure accesses the reference.

kinetiknz:
> What is the reasoning behind changes in [37fd803](https://github.com/mozilla/cubeb-coreaudio-rs/commit/37fd803332f4e13264ed26a7e81f279cbf2476b8)? It seems wrong to me because the lifetime of context is tied to the underlying queue and not the guard. This is especially evident in the `run_async` since there the guard is usually dropped before the closure accesses the reference.

You're right, thanks.  I've reverted my change and applied yours.  There's still an issue with the `Queue` wrapper that allows [d9ebd82](https://github.com/mozilla/cubeb-coreaudio-rs/pull/253/commits/d9ebd823aac4854e8c1991bc6a826b852d896e60) to happen, but I'll look at fixing that separately.

PR #254, opened 2025-07-09
Call the device change callback when the default device changes

padenot:
Does what it says. I had to write some workarounds, to be investigate later, for tests to pass on my new mac.


padenot:
The same changes in other backends: https://github.com/mozilla/cubeb/pull/819

padenot:
Yeah, 14 and 15 are a bit touchy still, it seems like.

src/backend/tests/api.rs
```diff
@@ -1165,6 +1218,11 @@ fn test_get_channel_count_of_input_devices_with_vpio() {
 
     let vpio_channel_counts = run_serially_forward_panics(get_nonvpio_input_channel_counts);
     assert_eq!(non_vpio_channel_counts, vpio_channel_counts);
```
ChunMinChang: duplicate check w/ the one in the if block. 

This check fails without the changes made here on my MBP w/ M1 Max, so it doesn't appear to be an M4-specific issue. I've open #256 tracking this.

src/backend/tests/api.rs
```diff
@@ -1998,3 +2060,192 @@ fn test_shared_voice_processing_release_on_idle_cancel_on_take() {
...
+
+    for (scope, device_type) in scope_types {
+        let (_initial_default, alternate_device) = device_info.get(scope).unwrap();
+        let scope_name = match scope {
+            Scope::Input => "input",
+            Scope::Output => "output",
+        };
+
+        // Change device
+        let output = Command::new("SwitchAudioSource")
```
ChunMinChang: This fails on my MBP because I don't have `SwitchAudioSource` installed.

Is there a specific reason for not using `test_set_default_device` instead?

From what I can tell, `SwitchAudioSource` sets the default device in a similar way to `test_set_default_device` (see [source](https://github.com/deweller/switchaudio-osx/blob/6659247d96a6b1f0c6b89bfd33a597ed36030c9a/audio_switch.c#L586)). Is there a particular reason for choosing `SwitchAudioSource` over `test_set_default_device`?

src/backend/tests/api.rs
```diff
@@ -1998,3 +2060,192 @@ fn test_shared_voice_processing_release_on_idle_cancel_on_take() {
     let r2 = queue.run_sync(|| shared.take()).unwrap();
     assert!(r2.is_ok());
 }
+
```
ChunMinChang: This should be moved to *device_change.rs*, as all tests related to device changes are located there. Additionally, the following tests should be included in *run_device_tests.sh* so they can be automatically run in CI.

src/backend/tests/api.rs
```diff
@@ -1998,3 +2060,192 @@ fn test_shared_voice_processing_release_on_idle_cancel_on_take() {
...
+            Scope::Output => "output",
+        };
+
+        Command::new("SwitchAudioSource")
+            .args(&["-t", scope_name, "-i", &initial_default.to_string()])
+            .output()
+            .expect("Failed to restore default device");
+    }
+}
+
```
ChunMinChang: These tests should be included in *run_device_tests.sh* with something like:
```
cargo test test_device_collection_change_on_default_input_device_change -- --ignored --nocapture

...
```

src/backend/tests/api.rs
```diff
@@ -1998,3 +2060,192 @@ fn test_shared_voice_processing_release_on_idle_cancel_on_take() {
...
+    // Register callbacks for all requested device types
+    for (_, device_type) in scope_types {
+        let callback_fn = match *device_type {
+            DeviceType::INPUT => input_device_collection_changed_callback,
+            DeviceType::OUTPUT => output_device_collection_changed_callback,
+            _ => panic!("Unsupported device type: {:?}", device_type),
+        };
+
+        assert!(context
+            .register_device_collection_changed(*device_type, Some(callback_fn), callback_data_ptr,)
```
ChunMinChang: wandering `,` at the end

src/backend/tests/api.rs
```diff
@@ -1998,3 +2060,192 @@ fn test_shared_voice_processing_release_on_idle_cancel_on_take() {
...
+            .is_ok());
+    }
+
+    // Change default devices and wait for callbacks
```
ChunMinChang: `Notifier` with`Watcher` should be able to do this job. See `test_plug_and_unplug_device_in_scope` for more details.

src/backend/mod.rs
```diff
@@ -2560,6 +2615,73 @@ impl AudioUnitContext {
...
+            );
+            let ret = audio_object_add_property_listener(
+                kAudioObjectSystemObject,
+                &input_address,
+                audiounit_default_device_changed_callback,
+                context_ptr,
+            );
+            if ret != NO_ERR {
+                // Clean up the hardware devices listener on failure
+                let address = get_property_address(
```
ChunMinChang: Thr declaration for `address` can be saved since it's the same as above one

src/backend/mod.rs
```diff
@@ -2560,6 +2615,73 @@ impl AudioUnitContext {
                 );
                 return Err(Error::error());
             }
+
+            // Also listen for default device changes
+            let input_address = get_property_address(
```
ChunMinChang: `default_input_address` sounds clearer. The `address` above should be renamed to something like `hardware_address` as well.

src/backend/mod.rs
```diff
@@ -2560,6 +2615,73 @@ impl AudioUnitContext {
...
+            );
+            let ret = audio_object_add_property_listener(
+                kAudioObjectSystemObject,
+                &output_address,
+                audiounit_default_device_changed_callback,
+                context_ptr,
+            );
+            if ret != NO_ERR {
+                // Clean up both previous listeners on failure
+                let address = get_property_address(
```
ChunMinChang: same as above

src/backend/mod.rs
```diff
@@ -2560,6 +2615,73 @@ impl AudioUnitContext {
...
+                    &input_address,
+                    audiounit_default_device_changed_callback,
+                    context_ptr,
+                );
+                cubeb_log!(
+                    "Cannot add default output device listener for {:?}, Error: {}",
+                    devtype,
+                    ret
+                );
+                return Err(Error::error());
```
ChunMinChang, line 2690: It would be cleaner to place this cleanup code inside a `finally` block or or a closure. Since the return values from `audio_object_remove_property_listener` are ignored, you can call it with all the address values in one place to unregister the callbacks, regardless of whether they were actually registered.

src/backend/mod.rs
```diff
@@ -2610,22 +2732,37 @@ impl AudioUnitContext {
...
+            DeviceType::INPUT | DeviceType::OUTPUT,
+        );
+        audio_object_remove_property_listener(
+            kAudioObjectSystemObject,
+            &output_address,
+            audiounit_default_device_changed_callback,
+            context_ptr,
+        );
+
+        Ok(())
```
ChunMinChang: No `Err(Error::error())` if one of the `audio_object_remove_property_listener` fails?

ChunMinChang (commented):
Is this for https://bugzilla.mozilla.org/show_bug.cgi?id=1879204? If it is, could you brief the motivation behind this change in the commit message or the PR comment?

Otherwise, could you provide more context on what this change is intended to address? I didn't see a bug number linked in the PR, nor could I find any related changes in the cubeb or cubeb-pulse repositories.

src/backend/tests/api.rs
padenot: It's strange, I don't remember it failing on my M1 Max, but maybe it is more about OS update vs. hardware change (I did both at the same time).

padenot:
@ChunMinChang I've addressed the comments in another commit so it's easier for you to review the changes, but I'll squash before merging. Thanks!

ChunMinChang:
LGTM thanks!

ChunMinChang (approved):

PR #265, opened 2026-01-07
Claude settings

ChunMinChang:
experiment

### DIFF
--- a/.claude/hooks/clippy-fix.sh
+++ b/.claude/hooks/clippy-fix.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+set -e
+
+cd "$CLAUDE_PROJECT_DIR" || exit 1
+
+TOOL_INPUT=$(cat)
+FILE_PATH=$(echo "$TOOL_INPUT" | jq -r '.tool_input.file_path // empty' 2>/dev/null || echo "")
+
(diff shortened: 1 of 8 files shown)


mutexlox-signal:
(just saw this in my inbox as I'm watching this repo)

I would personally tread **extremely** carefully here; in my experience (though I haven't used claude specifically), LLMs tend to produce subtly wrong output, particularly on highly specialized areas such as audio processing code that are not as commonly represented in training corpora. Further, they tend to be wrong in ways that are hard to notice without extremely careful attention, often requiring as much work as just implementing the fix by hand to begin with.

For example, I recently ran across a (possibly) LLM-generated website attempting to answer a subtle question about the specific wire format of packets transmitted when in Opus's DTX mode that conflated an obsolete mode of operation with the current behavior, producing what I believe was a description of a system that actually never was in existence, and it took careful attention to the opus RFC to realize that that had happened.

ChunMinChang:
Thanks for the insight! I am aware of the misinformation Claude could provide, hence the changes here only add basic settings like hooks that will run automatically to format the code and fix the lint issues. There is also a script blocking Claude to access some sensitive data like .ssh or some system configurations.

However, I also found Claude and other AI tools can simplify the workflow in many situations, and it's good enough to do some repetitive works. Additionally, for me, it's a good companion for brainstorming.

I'd like to leave this option open for now, to see what you and all others think, and will add comments to record what I found.

PR #274, opened 2026-03-13
Add TSan annotations to avoid false-positive TSan failures

ChunMinChang:
ThreadSanitizer reports false-positive data races in `test_ops_duplex_voice_stream_set_input_processing_params*` because it cannot observe CoreAudio's internal synchronization. `AudioOutputUnitStop `waits for in-flight callbacks via an internal `HALB_Mutex`, establishing a happens-before relationship, but this is invisible to TSan since it occurs inside Apple's closed-source framework.

This PR teaches TSan about CoreAudio's internal synchronization by adding TSan annotations (gated behind a `tsan-annotations` feature flag), and run TSan in two passes so annotations are only applied to known false-positive tests while maximizing unmodified TSan coverage on everything else.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,6 +50,10 @@ jobs:
     - name: Regular Test
       run: sh run_tests.sh
     
-    - name: Sanitizer Test
+    - name: AddressSanitizer Test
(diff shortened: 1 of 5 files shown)


kinetiknz (approved):
LGTM, thanks!

PR #275, opened 2026-03-25
Handle zero rate/channels from hardware description gracefully

kinetiknz:
CoreAudio can report zero sample rate or channel count in the hardware stream description during device transitions (e.g. device unplug, default device change).  This is the #1 cubeb-coreaudio crash signature, with ~240 crash pings per 30 days on macOS, 100% in the main (parent) process.

Previously, create_stream_description() would panic on zero rate or channels.  Now:
- Validate rate and channel count in setup() immediately after reading the hardware description, returning an error with diagnostic logging when the values are invalid.  This catches the problem at the source, where we have full context (device info, stream pointer).
- Convert the asserts in create_stream_description() to debug_assert + error return, so programming errors are still caught in dev builds but users don't crash.
- The output path already had a zero-channel check ("observed in the wild"); extend it to also check rate, and add both checks to the input path.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -267,8 +267,22 @@ fn create_device_info(devid: AudioDeviceID, devtype: DeviceType) -> Option<devic
 }
 
 fn create_stream_description(stream_params: &StreamParams) -> Result<AudioStreamBasicDescription> {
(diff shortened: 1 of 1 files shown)


ChunMinChang (approved):
LGTM, thanks!

PR #276, opened 2026-03-25
Handle null AudioUnit gracefully after failed reinit

kinetiknz:
After a failed device reinit, AudioUnits are closed (nulled) and the stream enters an error state, but calls to set_volume or set_input_processing_params can still arrive before the client processes the error notification.  These calls are dispatched to the serial queue, where they encounter a null AudioUnit and crash the process (~95 crash pings per 30 days on macOS).

Fix:
- Convert assert!(!unit.is_null()) to error returns with logging in set_volume, get_volume, set_input_mute, and set_input_processing_params.  Retain debug_assert so unexpected null units are still caught in dev builds.
- Set stopped=true in the reinit_async error path so the stream is in a consistent non-operational state after failed reinit, preventing callbacks from running on a closed stream.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -309,7 +309,11 @@ fn create_stream_description(stream_params: &StreamParams) -> Result<AudioStream
 }
 
 fn set_volume(unit: AudioUnit, volume: f32) -> Result<()> {
-    assert!(!unit.is_null());
+    debug_assert!(!unit.is_null());
+    if unit.is_null() {
+        cubeb_log!("set_volume: audio unit is null");
+        return Err(Error::Error);
+    }
     let r = audio_unit_set_parameter(
         unit,
         kHALOutputParam_Volume,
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -309,7 +309,11 @@ fn create_stream_description(stream_params: &StreamParams) -> Result<AudioStream
 }
 
 fn set_volume(unit: AudioUnit, volume: f32) -> Result<()> {
-    assert!(!unit.is_null());
+    debug_assert!(!unit.is_null());
+    if unit.is_null() {
```
ChunMinChang, line 313: It's not easy to tell why both assertion and if-check for non-null both exist here without checking the commit message or PR, mind adding some comment somewhere?

This situation makes me wonder what else APIs we shouldn't call after the stream runs into the error state. We don't have a state variable indicating the stream status now, but maybe we should? That would be in another PR I think.

ChunMinChang (approved):
LGTM, thanks!

PR #277, opened 2026-03-25
Handle zero-frame input callbacks and output buffer size mismatches

kinetiknz:
CoreAudio can call the input callback with 0 frames during device transitions.  The output callback already handles this case (added for the same reason), but the input callback had an assert that crashed the process (~21 crash pings per 30 days on macOS).  Add the same early-return pattern.

Similarly, during device format changes there can be a transient mismatch between the output buffer size provided by CoreAudio and the format described by output_dev_desc.  The mixer assert would crash if the buffer was too small (~19 crash pings per 30 days).  Replace with a check that outputs silence and returns early, avoiding an out-of-bounds write in the mixer.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -515,11 +515,18 @@ extern "C" fn audiounit_input_callback(
         Reinit,
     }
 
-    assert!(input_frames > 0);
     assert_eq!(bus, AU_IN_BUS);
 
     assert!(!user_ptr.is_null());
     let stm = unsafe { &mut *(user_ptr as *mut AudioUnitStream) };
+
+    if input_frames == 0 {
+        cubeb_alog!(
+            "({:p}) input callback empty.",
+            stm as *const AudioUnitStream
+        );
+        return NO_ERR;
+    }
 
     if unsafe { *flags | kAudioTimeStampHostTimeValid } != 0 {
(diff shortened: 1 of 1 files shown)


src/backend/mod.rs
```diff
@@ -980,10 +987,17 @@ extern "C" fn audiounit_output_callback(
 
     // Mixing
     if let Some(mixer) = stm.core_stream_data.mixer.as_mut() {
-        assert!(
-            buffers[0].mDataByteSize
-                >= stm.core_stream_data.output_dev_desc.mBytesPerFrame * output_frames
-        );
+        let needed = stm.core_stream_data.output_dev_desc.mBytesPerFrame * output_frames;
+        if buffers[0].mDataByteSize < needed {
```
ChunMinChang, line 991: nit: After checking the code, I find we always use `buffers[0]`. Since we have a `assert_eq!(out_buffer_list_ref.mNumberBuffers, 1)` check, maybe we should create a `AudioBuffer` variable referencing to `buffers[0]`. Otherwise my brain always wonder: why we only take care of the first item in `buffers` when reading the code. But it should be done in another PR I think.

ChunMinChang (approved):
LGTM, thanks!

PR #278, opened 2026-03-25
Handle unknown device IDs and empty aggregate devices gracefully

kinetiknz:
During device enumeration, devices can disappear between the list query and the property query, resulting in
kAudioObjectUnknown being passed to get_device_uid and get_channel_count.  These asserts crash the process (~48 crash pings per 30 days on macOS).  Replace with error returns using kAudioHardwareBadObjectError, which callers already handle.

Also convert the assert!(size > 0) in
activate_clock_drift_compensation to debug_assert.  When size is 0, the existing subdevices_num < 2 check immediately below returns the appropriate error, so the assert is redundant but was crashing the process (~40 crash pings per 30 days).

### DIFF
--- a/src/backend/device_property.rs
+++ b/src/backend/device_property.rs
@@ -4,7 +4,11 @@ pub fn get_device_uid(
     id: AudioDeviceID,
     devtype: DeviceType,
 ) -> std::result::Result<StringRef, OSStatus> {
-    assert_ne!(id, kAudioObjectUnknown);
+    debug_assert_ne!(id, kAudioObjectUnknown);
+    if id == kAudioObjectUnknown {
+        cubeb_log!("get_device_uid: device id is kAudioObjectUnknown");
+        return Err(kAudioHardwareBadObjectError as OSStatus);
+    }
     debug_assert_running_serially();
 
     let address = get_property_address(Property::DeviceUID, devtype);
--- a/src/backend/aggregate_device.rs
+++ b/src/backend/aggregate_device.rs
@@ -565,7 +565,7 @@ impl AggregateDevice {
         if status != NO_ERR {
             return Err(Error::from(status));
         }
-        assert!(size > 0);
+        debug_assert!(size > 0);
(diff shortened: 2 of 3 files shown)


src/backend/device_property.rs
```diff
@@ -4,7 +4,11 @@ pub fn get_device_uid(
     id: AudioDeviceID,
     devtype: DeviceType,
 ) -> std::result::Result<StringRef, OSStatus> {
-    assert_ne!(id, kAudioObjectUnknown);
+    debug_assert_ne!(id, kAudioObjectUnknown);
```
ChunMinChang, line 7: I am a bit confused about the timing this happens, and reminds me the [BMO 1628411](https://bugzilla.mozilla.org/show_bug.cgi?id=1628411), which filters `kAudioObjectUnknown` out from the device list in `audiounit_get_devices_of_type`.

Is the timing between `audiounit_get_devices_of_type` and `create_cubeb_device_info` during `enumerate_devices` qury? (otherwise `audiounit_get_devices_of_type` should exclude the id from the list).

If this can happen any time during `create_cubeb_device_info` call, shouldn't we add this check in all the property queries, in addition to `get_channel_count` and `get_device_uid`?

Assume the answer above is yes, then, subsequently, deallocation for the error case of `create_cubeb_device_info` is required (e.g., similar to what `destroy_cubeb_device_info` does). Supposed the device id suddenly become `kAudioObjectUnknown` between `get_device_group_id` and `get_device_label`. In this case, the `group_id` string has been allocated but would not be deallocated. We probably need a `let mut cleanup = finally(|| { /* something similar to destroy_cubeb_device_info */ })` in `create_cubeb_device_info`'s error case, and call `cleanup.dismiss()` when query succeeds. This deallocation issue could happen in other error cases as well, so it's good to have here or in another PR.

ChunMinChang (commented):
Generally looks good, but I'd like to understand when this could happen to make a more fair judgement.

PR #280, opened 2026-04-23
Expose device audio I/O workgroup via capi

kinetiknz:
Query kAudioDevicePropertyIOThreadOSWorkgroup for each stream's output and input device during setup() and cache the result in CoreStreamData. Add a C-ABI entry point, audiounit_stream_get_workgroup, returning a retained os_workgroup_t so auxiliary threads that cooperate on a stream's audio deadline (e.g. audioipc's cross-process callback dispatch thread) can join the workgroup.

Introduces a small RAII WorkGroup wrapper over os_workgroup_t to manage the +1 retain transferred from CoreAudio and from the capi.  Query failures are non-fatal: older OS versions and some virtual devices do not publish a workgroup.

### DIFF
--- a/src/backend/tests/interfaces.rs
+++ b/src/backend/tests/interfaces.rs
@@ -1944,3 +1944,45 @@ fn test_ops_stereo_input_duplex_voice_stream_stop() {
         },
     );
 }
+
+#[test]
+fn test_ops_stream_get_workgroup_null_stream() {
+    // NULL input returns NULL, no crash.
+    let wg = unsafe { crate::capi::audiounit_stream_get_workgroup(ptr::null_mut()) };
+    assert!(wg.is_null());
+}
+
+#[test]
+fn test_ops_stream_get_workgroup() {
+    use coreaudio_sys_utils::sys::os_release;
+
+    test_default_output_stream_operation("stream: get workgroup", |stream| {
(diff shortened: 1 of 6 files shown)


src/backend/tests/interfaces.rs
```diff
@@ -1944,3 +1944,45 @@ fn test_ops_stereo_input_duplex_voice_stream_stop() {
...
+        if populated {
+            assert!(
+                !wg.is_null(),
+                "workgroup present internally but capi returned null"
+            );
+
+            // Querying a second time should return a fresh retained reference,
+            // not the same pointer invalidated.  Both must be released.
+            let wg2 = unsafe { crate::capi::audiounit_stream_get_workgroup(stream) };
+            assert!(!wg2.is_null());
```
ChunMinChang, line 1977: Should `assert_eq!(wq, wq2)` be added here?
I am a bit confused about why we need two calls here without checking the two handle are same or not. `os_release` returns void, so nothing can be checked. Since internal ref-count for the Property::IOThreadOSWorkgroup handle is not detectable I guess, the ref-count imbalance can only be caught under ASAN build.

ChunMinChang (approved):
LGTM, thanks!

padenot (approved):

PR #281, opened 2026-04-23
audiounit: reset async logger thread id before starting audio units

kinetiknz:
Match the WASAPI, OpenSL and AAudio backends which call
`cubeb_async_log_reset_threads()` before spawning or starting the render
thread that will produce async log messages.

Without this, the `cubeb_async_logger` ring buffer's `assert_correct_thread`
fires in debug builds when `VERBOSE` logging is enabled on an active session
(e.g. turning on the WebRTC log profile during an active call) because the
CoreAudio I/O proc thread differs from whatever thread previously recorded
itself as the ring buffer's producer.

Requires `cubeb-sys` 0.34.1 for the corrected FFI declaration of
`cubeb_async_log_reset_threads`.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1909711

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -138,18 +138,18 @@ dependencies = [
 
 [[package]]
 name = "cubeb-backend"
-version = "0.32.0"
+version = "0.34.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


ChunMinChang (approved):
LGTM, thanks!

PR #282, opened 2026-06-04
Bump coreaudio-sys to 0.2.18

indutny-signal:
`dispatch_queue_create_with_target` is aliased as `_dispatch_queue_create_with_target$V2` on macOS, but prior to 0.2.18 `coreaudio-sys` mistakenly thought that it was not aliased resulting in App Store review rejections for any Mac apps that linked against cubeb.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -121,9 +121,9 @@ checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
 
 [[package]]
 name = "coreaudio-sys"
-version = "0.2.17"
+version = "0.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


indutny-signal:
This causes Mac App Store submission problems for apps that rely on `cubeb` since Apple Review process considers `_dispatch_queue_create_with_target` to be an "undocumented" API use.

simlay:
> Sadly coreaudio-sys is not actively maintained so even though I opened a PR for them too I don't think that the likelihood of them merging the change is high.

I accept fixes but not particularly big features to coreaudio-sys. With `coreaudio-rs`, we moved to use the objc2 bindings rather than bindgen. When I was going through the ecosystem, I somehow missed this project being a direct dependency on coreaudio-sys.

Anyway, [`coreaudio-sys` 0.2.18 is published](https://crates.io/crates/coreaudio-sys/0.2.18) and so a `cargo update` should fix things.

indutny-signal:
Force pushed the PR! Thank you @simlay !

PR #283, opened 2026-06-08
Provide a way to prevent inclusion of code that relies on private APIs

emir-signal:
Applications that rely on private APIs may and will get rejected by Apple. This change introduces a feature 'no-private-apis' that, when turned on, ensures that no private APIs are used. Currently, the only private API in the codebase seems to be AudioDeviceDuck, but in the future there may be more.

Caveat: when 'no-private-apis' is turned on, calls to disable audio device ducking will be removed, but no alternative method to disable ducking is provided.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ audio-dump = []
 vpio-forcelist = []
 gecko-in-tree = ["cubeb-backend/gecko-in-tree"]
 tsan-annotations = []
+no-private-apis = ["coreaudio-sys-utils/no-private-apis"]
 
 # Workaround for https://github.com/rust-lang/cargo/issues/6745 to allow this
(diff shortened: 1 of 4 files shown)


padenot:
Yes, thank you very much for the patch, we'll try to be careful when modifying (as it is sometimes a necessity), but don't hesitate to ping us or send more code should we make the mistake.

Pehrsons:
@emir-signal @indutny-signal note there's the [AVAudioVoiceProcessingOtherAudioDuckingConfiguration](https://developer.apple.com/documentation/avfaudio/avaudiovoiceprocessingotheraudioduckingconfiguration?language=objc) property on VPIO for *some* ducking control on macOS 14+, which you could add support for if you need to avoid some ducking in your app. You can't turn it off entirely like we do with `AudioDeviceDuck`, however.

PR #284, opened 2026-06-15
Dispose audio units asynchronously during stream teardown

spohlMozilla:
AudioComponentInstanceDispose makes a synchronous round-trip through coreaudiod to the active output device's HAL driver. When that device is slow to respond, the call can block for a long time. This was observed on macOS with a third-party virtual audio device selected as the default output: tearing down a stream stalled in AudioComponentInstanceDispose for an extended period.

In Gecko's remoted-cubeb configuration the audio server runs every client on a single shared RPC thread (with the cubeb context as a thread-local on that thread). A stalled disposal during one stream's teardown therefore blocks that thread, which blocks every other audio client behind it: new audio-IPC connection setup and content-process AudioContext creation both hang, freezing the browser UI. See https://bugzilla.mozilla.org/show_bug.cgi?id=2045209.

Defer the audio unit disposal to a background dispatch queue, but only during final stream teardown (destroy_internal). By that point the units have been uninitialized and detached from the stream, so they are solely owned and safe to dispose off the serial queue, and AudioComponentInstanceDispose does not touch the cubeb context. The shared VPIO unit is still never disposed here, and the reinit and error paths continue to dispose synchronously since they recreate or reuse the units. This does not change the device's behavior; it keeps a slow teardown from blocking the thread other clients depend on.

Testing: cargo build, cargo fmt, and cargo clippy are clean. The change was validated in a Firefox build (macOS gtest and mochitest-media green on try) and confirmed by the bug reporter to resolve the freezes on their machine.

### DIFF
--- a/coreaudio-sys-utils/src/audio_unit.rs
+++ b/coreaudio-sys-utils/src/audio_unit.rs
@@ -132,6 +132,33 @@ pub fn dispose_audio_unit(unit: AudioUnit) -> OSStatus {
     unsafe { AudioComponentInstanceDispose(unit) }
 }
 
+// Dispose an audio unit on a background queue instead of blocking the caller.
+//
+// AudioComponentInstanceDispose makes a synchronous round-trip to coreaudiod,
+// which can stall for a long time when the active output device is slow to
+// respond (e.g. a third-party virtual audio device). When the caller is the
+// shared AudioIPC server RPC thread, that stall blocks every audio client (e.g.
+// connection setup and AudioContext creation), freezing the UI
+// (https://bugzilla.mozilla.org/show_bug.cgi?id=2045209). Use this only during
+// final stream teardown, where `unit` has already been uninitialized and
+// detached from its stream, so it is solely owned here and safe to dispose off
+// the serial queue.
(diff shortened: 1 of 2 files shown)


spohlMozilla:
@kinetiknz could you review this? You are assigned to bug 1863925 (https://bugzilla.mozilla.org/show_bug.cgi?id=1863925), which is the same class of freeze: the audioipc server runs every client on one shared RPC thread, so a synchronous backend call during stream teardown blocks all of them, including CreateAudioIPCConnection.

Profiling in bug 2045209 (https://bugzilla.mozilla.org/show_bug.cgi?id=2045209) traced the macOS case to AudioComponentInstanceDispose stalling on a slow output device, and this PR defers that disposal so teardown stops blocking the shared thread. It is macOS-only so it does not directly fix the Windows freeze in 1863925, but it is the same root pattern, and your read on the approach would be valuable.

spohlMozilla:
@ChunMinChang Gentle review ping. Please let me know if there is anything that I can do to move this forward. Thanks!

ChunMinChang:
Sorry for the delay. I'll circle back to this once I come back from PTO.

